PR #9, opened 2020-02-22
AISDK-79: Add streaming client

Chanc3:
Looking for some feedback on the use of builders for this implementation. This client is a little complex than the api since it requires building the url using the customers content-type and parameters.

- [x] Finalize the integration test
- [x] Create unit tests for content-type builder
- [x] Create unit tests for streaming client
- [x] ~~Anything else?~~

It appears [retrofit does not support websockets](https://github.com/square/retrofit/issues/924) (developers have been waiting  but there is another library that was inspired by retrofit called [Scarlet](https://github.com/Tinder/Scarlet)


src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,173 @@
...
+      throws URISyntaxException {
+    this.accessToken = accessToken;
+    this.streamContentType = streamContentType;
+    this.metadata = metadata;
+    this.customVocabulary = customVocabulary;
+    this.client = client;
+    this.host = host;
+    this.filterProfanity = filterProfanity;
+
+    String content = buildContentString(streamContentType);
```
kennygea: Should we have a `WebsocketRequestBuilder` class that we pass in `StreamContentType` and all necessary parameters?

kennygea: Although for the Python SDK, we create the client with `AccessToken` and `ContentType`, and then pass in metadata/custom vocabulary/filter profanity after Client initialization. 

I guess it depends on whether we think `StreamingClient` here can be re-used. (i.e toggling a mic button) If so, then I think we should pass the latter 3 properties in separately.

src/test/java/revai/integration/StreamingTest.java
```diff
@@ -0,0 +1,74 @@
...
+      } catch (IOException e) {
+        throw new RuntimeException(e.getMessage());
+      }
+    } catch (IOException e) {
+      throw new RuntimeException(e.getMessage());
+    }
+    return fileByteArray;
+  }
+
+  public void streamAudioToServer(
```
amikofalvy:
```suggestion
  private void streamAudioToServer(
```

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public void setFilterProfanity(Boolean filterProfanity) {
+    this.filterProfanity = filterProfanity;
+  }
+
+  public void connect(WebSocketListener listener) throws URISyntaxException {
+    String completeUrl = buildURL() + buildContentString(streamContentType);
+    this.request = new Request.Builder().url(completeUrl).build();
+    webSocket = client.newWebSocket(request, listener);
+    client.dispatcher().executorService().shutdown();
```
kbridbur: This makes the okhttpclient non reusable correct? If so do you think its a problem?

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      return createContentTypeString(content);
+    }
+  }
+
+  private String buildURL() throws URISyntaxException {
+    URIBuilder uriBuilder = new URIBuilder();
+    uriBuilder.setScheme("wss");
+    if (host != null ) {
+      uriBuilder.setHost(host);
```
kbridbur: Is this so its easy for us to stream to non production environments?

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,144 @@
...
+    return filterProfanity;
+  }
+
+  public void setFilterProfanity(Boolean filterProfanity) {
+    this.filterProfanity = filterProfanity;
+  }
+
+  public void connect(WebSocketListener listener) throws URISyntaxException {
+    String completeUrl = buildURL() + buildContentString(streamContentType);
+    this.request = new Request.Builder().url(completeUrl).build();
```
kbridbur: Does this need to be a variable on the class? Can it just be a variable private to this method?

src/main/java/revai/StreamingClient.java
Chanc3: I didn't consider it being reusable. This makes sense and I've removed it from the method.

src/main/java/revai/StreamingClient.java
Chanc3: Yes. Similar to what we do for the other sdk's this will allow us to point at stage, test and feature branches.

src/main/java/revai/StreamingClient.java
Chanc3: Good catch!

src/main/java/revai/StreamContentType.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public Integer getChannels() {
+    return channels;
+  }
+
+  public void setChannels(Integer channels) {
+    this.channels = channels;
+  }
+
+  public static class Builder {
```
dkokotov: do we really need this builder class? it seems like `StreamContentType` is a very simple java class, having a builder for it seems overkilll

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,206 @@
...
+  /**
+   * Sends an HTTP request and upon authorization is upgraded to a websocket connection. Use
+   * websocket listener to handle websocket events.
+   *
+   * @param listener
+   * @see WebSocketListener
+   * @throws URISyntaxException
+   */
+  public void connect(WebSocketListener listener) throws URISyntaxException {
+    String completeUrl = buildURL() + buildContentString();
```
dkokotov: minor-ish - why not add the content string inside buildURL? (using the URLBuilder API)? it seems a bit surprising that, unlike the other parameters, you append this one after creating the initial URL string

src/test/java/revai/integration/StreamingTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    StreamingClient streamingClient = new StreamingClient(EnvHelper.getToken(), streamContentType);
+    streamingClient.setFilterProfanity(true);
+    streamingClient.setMetadata(testName.getMethodName());
+
+    File file = new File("./src/test/java/revai/resources/english_test.raw");
+    byte[] fileByteArray = readFileIntoByteArray(file);
+    int chunk = 8000;
+    StreamingClientListener streamingClientListener = new StreamingClientListener();
+    streamingClient.connect(streamingClientListener);
```
dkokotov: one question, why make the caller construct the listener instead of having the client construct and return it? would we ever expect the call connect with something other than a `StreamingClientListener`?

src/main/java/revai/StreamContentType.java
Chanc3: I guess I added it because I'm use to working with builders and it's just a personal preference. I will remove it.

src/main/java/revai/StreamingClient.java
Chanc3: I'm going to try and explain this as best I can so bear with me :) It might not go as smooth as I'm hoping lol.

The `URIBuilder` sets parameters based on key-value pairs. Each parameter automatically appends `&` before the pair in the URI. 

Because of the format of the content-type does not use the `&` delimiter to separate its pairs I cannot simply use the `setParameter()` method on it. The content-type string uses a semi-colon delimiter.

In order to use this I would need to construct the content-type string without the `content-type=` present and then use `setParameter("content_type", STRING)`.

```
String content = "audio/x-raw;layout=interleaved;rate=16000;format=S16LE;channels=1";
uriBuilder.setParameter("content_type", content)
```

Instead of doing it like above I felt like the content-type string having it's own builder like method would be clearer and easier to manage. Do you think I should change it? If it's ok to use what I have I can add the content-type appending in the return from the `buildURL()`

src/test/java/revai/integration/StreamingTest.java
Chanc3: The `StreamingClientListener` was constructed purely for the purpose of the integration test. It's located in the /integration folder so customers will not be using it, at least they shouldn't be. Each customer will need to create their own version of a listener that extends `okhttp3` `WebSocketListener`. The reason I don't construct one for them and return it is because the listener dictates how each event is handled. This might just be my inexperience showing here but I don't think we can pass in a defined `WebSocketListener` and still give the customer control over how they choose to handle each event.

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,206 @@
...
+  private String host;
+  private StreamContentType streamContentType;
+  private String metadata;
+  private String customVocabularyId;
+  private Boolean filterProfanity;
+
+  /**
+   * Constructs the streaming client that is used to establish a websocket connection with the
+   * Rev.ai server and stream audio data. The user access token can be generated on the website at
+   * https://www.Rev.ai/access_token. To establish a successful connection a valid StreamContentType
```
amikofalvy:
```suggestion
   * https://www.rev.ai/access_token. To establish a successful connection a valid StreamContentType
```

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,206 @@
...
+  public void setCustomVocabularyId(String customVocabularyId) {
+    this.customVocabularyId = customVocabularyId;
+  }
+
+  public Boolean getFilterProfanity() {
+    return filterProfanity;
+  }
+
+  /**
+   * Setting the profanity filter is optional. We filter approx. 600 profanities, which covers most
```
amikofalvy: I wouldn't include details on how the profanity filter works. These details can change and I don't want to update the SDK just because we need to change this comment.

amikofalvy: You can instead refer customers to the documentation page.

src/main/java/revai/StreamingClient.java
amikofalvy: That was well put, Chance.

I understand that you do not want to use a URIBuilder to build the content type because the elements of ContentType are not joined by the `&` char. 

you may want to consider putting the string concatenation inside the buildUrl.

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,206 @@
...
+    Request request = new Request.Builder().url(completeUrl).build();
+    webSocket = client.newWebSocket(request, listener);
+  }
+
+  /**
+   * Sends data over websocket in the form of a ByteString
+   *
+   * @param byteString
+   */
+  public void sendBytes(ByteString byteString) {
```
amikofalvy: In our other SDKs we use `sendAudioData`. Is there a reason you chose this name instead?

src/main/java/revai/StreamingClient.java
dkokotov: yes, a combo of what andrew suggested and `urlBuilder.setParameter("content_Type", [contentTypestring)` is what i was thinking of.

so inside buildURL you would have 

```
// other parameters
String contentType = buildContentTypeString(); // without the content_type= part
uriBuilder.setParameter("content_type", contentType);
```

src/test/java/revai/integration/StreamingTest.java
dkokotov: makes sense - i actually missed that the StreamingClientListener was in the test directory, my mistake.

I thought you were providing this StreamingClientListener for clients to monitor the different kind of messages (using the latches). but i think i misinterpreted how those latches ought to be used.

So what you said totally makes sense, but i do wonder if we want to provide the users of the SDK something a bit more high level. if the client has to implement WebSocketListener themselves, they have to know some details about our protocol, eg that hypotheses are sent as JSON using text messages, there is a special "connected" message, etc.

Its not a huge deal and we can of course document all this, but in our other SDKs (node, python) we take care of that inside the client, so the caller just receives connected and data events with the data already parsed out. i think it would be nice to do this as well here.

I think it could be simple, we just define our own listener interface, something like

```
public interface RevAiWebSocketLisneter {
    public onClose(..);
    public onError(..);
    public onConnected(..);
    public onHypothesis(Hypothesis hypo);
}
```

StreamingClient.connect would accept an instance of one of these, and internally it would have a WebSocketListener implementation that does the appropriate deserialization and routing to those methods.

Chanc3: This a great suggestion. In your example method you have a Hypothesis object. Would you suggest I deserialize the hypothesis json into a Hypothesis object and we give customers access to that? I do this in our selenium java tests so this shouldn't be an issue.

src/main/java/revai/StreamingClient.java
Chanc3: No real reason. I will change it

src/test/java/revai/integration/StreamingTest.java
Chanc3: @dkokotov @amikofalvy when time permits please take a look at the changes I made this is is all pretty low priority compared to the onprem stuff. I went ahead and implemented the interface dan suggested and updated the integration test to use this. It works quite well. 

The main concern I have with how it is currently is that passing back a Hypothesis object instead of a generic String will means that changes made on the api side could break the SDK by preventing deserialization (like if we added a new property or the format changed). Because of this it's my belief is that we should not be deserializing into a custom object like what I have but instead be returning strings or generic GSON objects. What are your opinions on this?

src/main/java/revai/StreamingClient.java
Chanc3: After some testing I found that the best option was to just add the concatenation in the return of the buildUrl. When using the `setParameter()` method is does some URL encoding which results in the following: 
```
content_type=%26audio%2Fx-raw%3Blayout%3Dinterleaved%3Brate%3D16000%3Bformat%3DS16LE%3Bchannels%3D1
```

It appears that when the parameters are passed in with this URL encoding the Rev.ai server doesn't decode it and results in a content-type error after the websocket is upgraded. The easiest way to prevent this was to use the simple concatenation.

src/main/java/revai/RevAiWebsocketListener.java
```diff
@@ -0,0 +1,17 @@
+package revai;
+
+import okhttp3.Response;
+import revai.models.streaming.Hypothesis;
+
+public interface RevAiWebsocketListener {
```
dkokotov: lets add javadoc to document the interface and methods

src/main/java/revai/StreamContentType.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public String getContentType() {
+    return contentType;
+  }
+
+  public void setContentType(String content_type) {
+    this.contentType = content_type;
+  }
+
+  public String getLayout() {
```
dkokotov, line 36: lets add javadoc to explain what these properties are

src/main/java/revai/StreamContentType.java
```diff
@@ -0,0 +1,58 @@
...
+    this.rate = null;
+    this.format = null;
+    this.channels = null;
+  }
+
+  public String getContentType() {
+    return contentType;
+  }
+
+  public void setContentType(String content_type) {
```
dkokotov: minor - should be `contentType` not `content_type`

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /** Will close the websocket connection by informing the server that it's the End of Stream */
+  public void close() {
+    webSocket.send("EOS");
+  }
+
+  private String buildURL() throws URISyntaxException {
+    URIBuilder uriBuilder = new URIBuilder();
+    if (scheme != null) {
```
dkokotov, line 170: it would be simpler i think to default scheme to `wss`. e.g you would declare it as `private string scheme = "wss"`, but it can be overwritten with the setter. Ditto for `host`. then you don't need these `if`s here.

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+   * This method is used for internal testing of the websocket connection against a mock webserver
+   * and should not be used
+   *
+   * @param scheme
+   */
+  public void setScheme(String scheme) {
+    this.scheme = scheme;
+  }
+
+  public String getHost() {
```
dkokotov: do we need this? don't think its used anywhere

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+    return host;
+  }
+
+  /**
+   * This method is used for internal testing and should not be used
```
dkokotov: fwiw there may be a legitimate reason reason why someone would point at an alternative hostname, say they were testing a development version. of the api. so i think we don't have to say its only for internal testing

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+   */
+  public void setFilterProfanity(Boolean filterProfanity) {
+    this.filterProfanity = filterProfanity;
+  }
+
+  /**
+   * Sends an HTTP request and upon authorization is upgraded to a websocket connection. Use
+   * websocket listener to handle websocket events.
+   *
+   * @param revAiWebsocketListener
```
dkokotov: lets make sure to fill in any missing javadoc

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+    if (streamContentType.getChannels() != null) {
+      content.add("channels" + "=" + streamContentType.getChannels());
+    }
+    return content;
+  }
+
+  private String createContentTypeParameter(List<String> content) {
+    String completeContentType = content.stream().collect(Collectors.joining(";"));
+    return "&" + completeContentType;
```
dkokotov: super minor: i think you can just do `String.join(";", content)`, slightly simpler.
also minor / pedantic, but i would return it without the `&contentType=` - instead in buildUrl I would be 
`return uriBuilder.build().toString() + "&content_type=buildContentString()`.

The reason is - the value of the content type is "[type];layout=foo;rate=x" etc. then you are passing that as the value of the `content_type` query parameter when building the url. so it makes sense to keep that separation. In fact i would move the construction of the content type string to a method on `StreamContentType`, similar as in the node SDK (https://github.com/revdotcom/revai-node-sdk/blob/401af632ec3e89c6051b3ff9812963c0b484d5fe/src/models/streaming/AudioConfig.ts#L32)

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+    webSocket.send(byteString);
+  }
+
+  /** Will close the websocket connection by informing the server that it's the End of Stream */
+  public void close() {
+    webSocket.send("EOS");
+  }
+
+  private String buildURL() throws URISyntaxException {
+    URIBuilder uriBuilder = new URIBuilder();
```
dkokotov: not a huge deal, but it looks like we are taking dependency on org.apache.httpclient library just for this url builder. 
OkHttp already has a url builder: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-http-url/. so i think its better to use that and avoid an extra dependency.

reducing number of dependencies is generally good practice, but httpclient is particularly thorny because often it is included in an appserver like Tomcat itself, and there can be conflict if its also included with an application WAR. i actually ran into this when using your library with Red5 server.

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Setting metadata is optional but helps to identify the stream.
+   *
+   * @param metadata String size restricted to <= 256 characters
+   */
+  public void setMetadata(String metadata) {
+    this.metadata = metadata;
+  }
+
+  public String getCustomVocabularyId() {
```
dkokotov: two questions:

- why are these various properties (metadata, vocabulary id, etc) provided as setters, whereas the contentType is provided in the constructor?
- and taking it a step further, why set them on the client at all instead of providing them as parameters of `connect` method.

all of them i think are just to construct the URL when connecting the socket, so i dont think they need to be stored as properties of the client object. 

in part i guess it depends on how you expect to use the client. if you expect to reuse it for multiple connections, then it would make sense i think to pass all those configuration options to the connect method, since they might differ between connections.

This is how node SDK does it (https://github.com/revdotcom/revai-node-sdk/blob/develop/src/streaming-client.ts#L65), though there for some reason content type is still specified as constructor param, which I am not sure why that is (cc: @kbridbur)

src/main/java/revai/Listener.java
```diff
@@ -0,0 +1,49 @@
+package revai;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import okhttp3.Response;
+import okhttp3.WebSocket;
+import okhttp3.WebSocketListener;
+import revai.models.streaming.Hypothesis;
+
+public class Listener extends WebSocketListener {
```
dkokotov: does it need to be a public class?

with the RevAiWebsocketListener interface, this is not intended to be used by  users of the library, i think, its only used internally in the client. so i would make it a private nested class inside StreamingClient

src/test/java/revai/integration/ClientListener.java
```diff
@@ -0,0 +1,82 @@
...
+public class ClientListener implements RevAiWebsocketListener {
+
+  private CountDownLatch connectedLatch = new CountDownLatch(1);
+  private CountDownLatch closeLatch = new CountDownLatch(1);
+  private CountDownLatch finalHypothesisLatch = new CountDownLatch(1);
+  private Gson gson = new Gson();
+  private JsonObject connectedMessage;
+  private List<Hypothesis> partialHypotheses = new ArrayList<>();
+  private List<Hypothesis> finalHypotheses = new ArrayList<>();
+  private static final Logger LOG = LoggerFactory.getLogger(ClientListener.class);
```
dkokotov: do we need this logging in the released version of the libary.

My concern is (1) it adds another dependency, and (2) we don't document how to configure the logging system, so its unclear where the log messages would go. probably by default they would go to the console, which the users may not want.

it seems like it was useful for debugging, but i would take it out before release (along with the dependencies in the POM)

src/test/java/revai/unit/MockClientListener.java
```diff
@@ -0,0 +1,32 @@
+package revai.unit;
+
+import okhttp3.Response;
+import revai.RevAiWebsocketListener;
+import revai.models.streaming.Hypothesis;
+
+public class MockClientListener implements RevAiWebsocketListener {
```
dkokotov: minor, do you need this? cant you just mock it with mockito (which i think is already in your dependency list). eg you would just do `mock(RevAiWebsocketListener.class)` to get a mock instance.

src/main/java/revai/StreamingClient.java
Chanc3: I agree that it would be but I need this so I can run the unit tests against a fake/mock webserver. This was the simplest way I could find to unit test streaming.

src/main/java/revai/StreamingClient.java
Chanc3: I think this is a valid point. My reasoning behind this was mostly influenced by java's inability to make methods with optional parameters. I can easily overload that method to handle optional parameters.

src/main/java/revai/StreamingClient.java
dkokotov: but why would the unit tests be an issue. What I was suggesting was 

```
public class StreamingClient {
   private String scheme = "wss";

   public void setScheme(String scheme) {
      this.scheme = scheme;
   }

   private String buildURL() ... {
        uriBuilder.scheme(scheme).host(..)... // etc
    }
}
```

and similar for host. i think this would all still work just fine for your mock webserver, no?
its a very minor point of course.

src/main/java/revai/StreamingClient.java
dkokotov: yeah its annoying there is no default parameter values.

src/main/java/revai/StreamingClient.java
Chanc3: As discussed in slack I removed the `URIBuilder` and instead am using the `HttpUrl.Builder` supplied by OkHttp. The builder silently converts `https` and `http` schemes to `wss` and `ws`. Since the HttpUrl.Builder throws an error if the scheme does not match http or https I thought we could leave the error handling to the builder.

src/main/java/revai/RevAiWebSocketListener.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Listens for events over the WebSocket connection to Rev.ai
+ */
+public interface RevAiWebSocketListener {
+
+  /**
+   * Supplies the connection message received from Rev.ai.
+   *
+   * @param message a String in JSON format that contains the message type and job ID.
```
dkokotov, line 13: would make sense to parse this into a java class as well, eg it would be `onConnected(ConnectedMessage message)` where `ConnectedMessage` would have id/type properties (ConnectedMessage and Hypothesis could inherit from `StreamingResponseMessage base class which would have the type property)

src/main/java/revai/RevAiWebSocketListener.java
```diff
@@ -0,0 +1,50 @@
...
+   * @param hypothesis the partial or final hypothesis of the audio.
+   * @see Hypothesis
+   */
+  void onHypothesis(Hypothesis hypothesis);
+
+  /**
+   * Supplies the error and response received during a WebSocket ErrorEvent.
+   *
+   * @param t the error thrown.
+   * @param response the WebSocket response to the error.
```
dkokotov, line 31: minor grammar. - "the Webscocket response containing the error event"

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,300 @@
...
+   * @param metadata used to identify the stream.
+   * @param streamContentType content-type query parameter.
+   * @param customVocabularyId customVocabularyId the ID of the custom vocabulary.
+   * @param filterProfanity this is false by default.
+   * @see RevAiWebSocketListener
+   * @see StreamContentType
+   */
+  public void connect(
+      RevAiWebSocketListener revAiWebSocketListener,
+      String metadata,
```
dkokotov: I would put metadata, custom vocab id, and filter profanity into a `SessionConfig` class. then we don't need so many overloads, just two: one that takes only the content type, and one that takes content type and sessionConfig. all fields of this sessionConfig would be optional.

so eg one way to call would be 

```
connect(listener,
    new StreamContentType. { ContentType = "audio/mp3" }, 
    new SessionConfig { filterProfanity = true }
);
```

src/main/java/revai/StreamingClient.java
dkokotov: I would do it slightly differently, right now if you do `setScheme("ws")` it wouldnt work, which would be surprising - the fact that we need to use http/https behind the scenes shouldnt be exposed to the caller of our class. Even though its somewhat minor, since we dont normally expect people to change it at all.

simplest way i think would be to validate in the setter. something like this:

```
void setScheme(string scheme) {
    string actualScheme = scheme == "wss" ? "https" : (scheme == "ws" ? "http" : scheme);
    if (actualScheme != "http" && actualScheme != "https") { 
       throw new IllegalArgumentException("invalid scheme...");
    }. 
  
}
```

src/main/java/revai/StreamingClient.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+    if (host != null) {
+      urlBuilder.host(host);
+    } else {
+      urlBuilder.host("api.rev.ai");
+    }
+    if (port != null) {
+      urlBuilder.port(port);
+    }
+    urlBuilder.addPathSegment("speechtotext");
```
dkokotov: super minor, i think you could also do `addPathSegments("speechtotext/v1/stream")` to be a bit less verbose. (note the different method name)

pom.xml
```diff
@@ -31,12 +31,6 @@
             <version>3.2.4</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.maven</groupId>
```
dkokotov, line 35: pls remove the apache.httpclient dependency as well.

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>revai</groupId>
     <artifactId>revai-java-sdk</artifactId>
-    <version>0.0.0</version>
+    <version>0.0.1</version>
```
dkokotov: we will make it 1.0.0 before we release the first public version right?

Chanc3: Yes this was just left over from me creating a jar for testing the ApiClient. Should I set it to 1.0.0 now?

src/main/java/revai/StreamingClient.java
Chanc3: This makes sense. I think conditional expressions can be a little obscure. Do you have any objections to using a switch statement instead? Like this...
```
public void setScheme(String scheme) {
    switch (scheme.toLowerCase()) {
      case "wss":
      case "https":
        this.scheme = "https";
        break;
      case "ws":
      case "http":
        this.scheme = "http";
        break;
      default:
        throw new IllegalArgumentException("Invalid scheme: " + scheme);
    }
  }
```

pom.xml
Chanc3: yup that one is gone too :)

pom.xml
dkokotov: sure.

PR #11, opened 2020-04-08
AISDK-94: Setup travis CI machines

Chanc3:
The sdk project has not yet been configured to send requests multiple environments like `api-test` or `api-stage`. That configuration / setup will be addressed in a later ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: java
+
+jdk:
+  - oraclejdk8
+  - oraclejdk9
+  - oraclejdk10
+  - oraclejdk11
+
+# see https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure
+sudo: false
+
+env:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,27 @@
+language: java
+
+jdk:
+  - openjdk8
```
kbridbur, line 4: Are we sure we want to support all these versions?

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+  - oraclejdk11
+  - oraclejdk12
+
+# see https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure
+sudo: false
+
+env:
+  - secure: "GOnKWQXhSe9ksS4/mlRvXR+CB4nLTYUm5w4opz61hoeDDkASSlxKYi//5kVGL2hSke2owdDXxjaWpsjrqFHAG1JLkwtNoY/I9NfykqxZyBtXkQlksRckknMiCnfpJu6biDTLAobk6XRQv/j6LUYjtrgLNM3PrjIQxv86AmCAw4vOIEAoR0sS4z9TrIJ45nyTROLAR4RQAQ16nF7HM3b6aqhoMaio44e0vxUr1404U69uIqJRCVmbPIstlXBgD1HqpXJ3XuKJtaayvZaha+qggc5bzLr1SfywvIcRaKJp2BOUF1uZHDRZuSDns0WKBCbyPGRHl5qLjCjx51h/fL7q9daycB+1Amv+6BJXqOhHXBa4D1rMPNLcdwn34oKDNwNqbA6w0C+zpsRD4h5dpRBSX1KBIccChQRG71bigsCxH4weMikjKT0NocWSursB55Ges/onL6SzmJiPa7ibeju7ifem/F1RbZHoj4Tt+3+yNm4CHFFV/k/9PtnaV8s86HKZZHb19HFtExDtLUXAlI8E+YXPWf4RrU5M0KPSRRbmI166t4zrB6YoFJJ36rKjkSbbNghuoo+/d9d204d5R/+tm9wXAXHuPlkArjVke2d1SnFO8rXizXt0WNWZfsryG5gTli4lx8vFpV12ixPzUfibUu7S0xWz6RCZDBw0nsf+ekQ="
+
+# cache the build tool's caches
```
kbridbur: This comment isn't super descriptive. I assume caches a lot of the jdk stuff that's required to run the tests? It would be nice if this comment was a bit more detailed

kbridbur (commented):
Looks pretty good :). A few questions

.travis.yml
Chanc3: After doing more research we only want to support the versions designated as Long-Term-Support (LTS) releases. This would mean Java 8 and 11. Oracle designates an LTS release every 3 years. Even though Java 14 is out it is not an LTS and so we don't have to worry about supporting it. The next LTS should be released in ~2021

.travis.yml
Chanc3: Updated to:
```
# cache the build dependencies into the Maven local repository
```

kbridbur (approved):
:shipit:

PR #13, opened 2020-04-17
AISDK-96: Add javadoc notation and populate README

Chanc3:
- [x] Add javadoc notation to public classes, methods and variables
- [x]  Populate README.md with usage details


README.md
```diff
@@ -2,6 +2,220 @@
...
 
+## Documentation
+
+See the [API docs](https://www.ai.rev.ai/docs) for more information about the API.
+
+## Install the SDK
+The recommended way to use the Rev.ai Java SDK is to import it into the project using Maven.
+
+      <dependency>
+        <groupId>ai.rev</groupId>
```
amikofalvy, line 13: The reverse domain format is a bit confusing with short domain names and non-dotcom domain suffixes.

As I understand it, this is a requirement for our groupId to match our domain name. Is this correct?

README.md
```diff
@@ -2,6 +2,220 @@
...
+### Submitting a job
+
+You can submit a local file
+```
+RevAiJob revAiJob = apiClient.submitJobLocalFile("./path/to/file.mp3");
+```
+
+or submit via a public url
+```
+RevAiJob revAiJob = apiClien.submitJobUrl("https://www.ai.ai.rev.ai/FTC_Sample_1.mp3");
```
amikofalvy:
```suggestion
RevAiJob revAiJob = apiClient.submitJobUrl("https://www.ai.ai.rev.ai/FTC_Sample_1.mp3");
```

README.md
```diff
@@ -2,6 +2,220 @@
...
+FileInputStream fileInputStream;
+try {
+  fileInputStream = new FileInputStream(file);
+} catch (FileNotFoundException e) {
+  throw new RuntimeException("Could not find file [" + file.getName() + "]");
+}
+RevAiJob revAiJob = apiClient.submitJobLocalFile(fileInputStream, String fileName, RevAiJobOptions options);
+```
+
+`RevAiJob` objects contain all the information normally found in a successful response from our
```
amikofalvy:
```suggestion
`RevAiJob` objects contain job information as defined by the documentation...
```

amikofalvy: I think you can omit the concept of success or failure here. We have a separate section for Errors.

README.md
```diff
@@ -2,6 +2,220 @@
...
+[Submit Job](https://www.ai.rev.ai/docs#operation/SubmitTranscriptionJob) endpoint.
+
+If you want to get fancy, all send job methods can take a `RevAiJobOptions` object which contains methods
```
amikofalvy: Use consistent terminology:
```suggestion
If you want to get fancy, all submit job methods have overrides that allow specifying `RevAiJobOptions` to configure job specific settings: `metadata`, ...
```

README.md
Chanc3: Yes and this follows maven groupid naming convention.
http://maven.apache.org/guides/mini/guide-naming-conventions.html

README.md
amikofalvy: I would also mention the list is non-exhaustive or reference a list of job options or the job options object definition instead of listing them out. We don't want to have to update this README's list every time we add a new job option.

README.md
```diff
@@ -2,6 +2,220 @@
...
+to set optional parameters such as: `metadata`, `callback_url`, `skip_diarization`,`skip_punctuation`, 
+`speaker_channels_count`, `filter_profanity`, `remove_disfluencies`  and `custom_vocabularies`. These are 
+also described in the request body of the [Submit Job](https://www.ai.rev.ai/docs#operation/SubmitTranscriptionJob) endpoint.
+
+### Checking your job's status
+
+You can check the status of your transcription job using its `id`
+
+```
+RevAiJob revAiJob = apiClient.getJobDetails(revAiJob.getJobID());
```
amikofalvy: This example contains a cyclical reference. Without context it doesn't make complete sense.

You could change the resulting revAiJob's name to `newlyRefreshedRevAiJob`

README.md
```diff
@@ -2,6 +2,220 @@
...
+```
+
+`RevAiJob` objects contain all information normally found in a successful response from
```
amikofalvy: Duplicate information from the previous. I would recommend my wording that I mention above.

README.md
```diff
@@ -2,6 +2,220 @@
...
+our [Get Job](https://www.ai.rev.ai/docs#operation/GetJobById) endpoint
+
+### Checking multiple files
+
+You can retrieve a list of transcription jobs with optional parameters
+
+```
+List<RevAiJob> jobs = apiClient.getListOfJobs();
```
amikofalvy, line 93: Minor -  why didn't we call the method name simply`getJobs()`?

README.md
```diff
@@ -2,6 +2,220 @@
...
+
+// limit amount of retrieved jobs
+List<RevAiJob> jobs = apiClient.getListOfJobs(3);
```
amikofalvy: use local variables to ensure clarity since java does not allow named parameters.

README.md
```diff
@@ -2,6 +2,220 @@
...
+
+// get jobs starting after a certain job id
+List<RevAiJob> jobs = apiClient.getListOfJobs("Umx5c6F7pH7r");
```
amikofalvy: ditto on local variable

README.md
```diff
@@ -2,9 +2,223 @@
...
+The recommended way to use the Rev.ai Java SDK is to import it into the project using Maven.
+
+      <dependency>
+        <groupId>ai.rev</groupId>
+        <artifactId>revai-java-sdk</artifactId>
+        <version>1.0.0</version>
+      </dependency>
+    
+## Installing it locally
+Once you've cloned the repo you can use Maven to build it locally and install it your local Maven .m2 repository.
```
jennywong2129: install it `in` your local Maven ...?

README.md
```diff
@@ -2,9 +2,223 @@
...
+### Submitting a job
+
+You can submit a local file
+```
+RevAiJob revAiJob = apiClient.submitJobLocalFile("./path/to/file.mp3");
+```
+
+or submit via a public url
+```
+RevAiJob revAiJob = apiClient.submitJobUrl("https://www.ai.ai.rev.ai/FTC_Sample_1.mp3");
```
jennywong2129: is that a real media url?

README.md
Chanc3: It's not too late 😉

Chanc3: But this does match the documentation https://www.rev.ai/docs#operation/GetListOfJobs

README.md
Chanc3: I used to be lol. It got messed up in a refactor. I thought I found all of those. Good catch!

README.md
amikofalvy: I'm not a fan of that wording in our documentation. I think GET /jobs implies that it will return a list of some kind.

amikofalvy: I don't want to change the wording if we use the same `getlistofjobs` in other sdks. The higher priority is being consistent.

README.md
```diff
@@ -2,9 +2,225 @@
...
+## Install the SDK
+The recommended way to use the Rev.ai Java SDK is to import it into the project using Maven.
+
+      <dependency>
+        <groupId>ai.rev</groupId>
+        <artifactId>revai-java-sdk</artifactId>
+        <version>1.0.0</version>
+      </dependency>
+    
+## Installing it locally
```
dkokotov: Maybe we should call this section "Build and install locally from source". Otherwise it's a bit unclear how it relates to the previous section (install the SDK).

README.md
```diff
@@ -2,9 +2,225 @@
...
+[RevAiJobOptions](./src/main/java/ai/rev/models/asynchronous/RevAiJobOptions.java) to configure job specific settings. These are also described in the 
+request body of the [Submit Job](https://www.ai.rev.ai/docs#operation/SubmitTranscriptionJob) 
+endpoint.
+
+### Checking your job's status
+
+You can check the status of your transcription job using its `id`
+
+```
+RevAiJob newlyRefreshedRevAiJob = apiClient.getJobDetails(revAiJob.getJobID());
```
dkokotov: small nitpick on the theme of naming, but could we use `JobId` instead of `JobID`everywhere (and in general Id vs ID). Generally we use Id in all of our other code.

amikofalvy (approved):
Looks good. I think this is some of the most complete documentation I've seen :)

It is curious how git identified many file renames/changes correctly but for a few files it marked them as deleted and new files.

:shipit:

README.md
```diff
@@ -2,9 +2,225 @@
...
+streamingClient.connect(clientListener, streamContentType, sessionConfig);
+```
+
+You can stream data over the WebSocket in the form of a `ByteString` using the `streamingClient.sendAudioData()` method.
+
+```
+streamingClient.sendAudioData(ByteString);
+```
+
+The streaming connection will close when you call the method `streamingClient.close()` or if you go 15 seconds without sending any audio data.
```
dkokotov, line 218: It would be good to provide guidance on how a StreamingClient instance should be reused (or not) across multiple streaming sessions (both simultaneous and not).

README.md
Chanc3: Yes this is consistent with the other SDKs as well

README.md
Chanc3: Good catch!

src/main/java/ai/rev/ErrorInterceptor.java
```diff
@@ -1,10 +1,16 @@
-package revai;
+package ai.rev;
```
dkokotov: if we are going to be doing this, I think we should have another top-level package, e.g. `ai.rev.speechtotext`. Since now we are using `ai.rev` as basically the company domain name prefix, and i don't think putting all the classes directly into that makes sense, if we ever have other APIs or libraries we could easily have name collisions.

src/main/java/ai/rev/StreamContentType.java
```diff
@@ -0,0 +1,170 @@
...
+public class StreamContentType {
+
+  private String contentType;
+  private String layout;
+  private Integer rate;
+  private String format;
+  private Integer channels;
+
+  public StreamContentType() {
+    this.contentType = null;
```
dkokotov: minor, but I dont think these initializations are necessary, null is the default value.

src/main/java/ai/rev/models/asynchronous/RevAiAccount.java
```diff
@@ -0,0 +1,55 @@
+package ai.rev.models.asynchronous;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * The RevAiAccount object provides basic information about a Rev.ai account associated with a valid
+ * access token.
+ *
+ * @see <a href="https://www.rev.ai/docs#tag/Account">https://www.rev.ai/docs#tag/Account</a>
```
dkokotov: pretty impressively comprehensive Javadoc!

src/main/java/ai/rev/ErrorInterceptor.java
Chanc3: shoot I didn't think of that... I'll reach out to Sonatype and see if we can add that to our group id. I imagine we can because we now own that name space in the central repo. Do we want `ai.rev.speechtotext`?

Chanc3: I'm going to make the change. I heard back from Sonatype and they grant permissions at the top level group id so it won't be an issue for this project.

dkokotov: cool, yeah `ai.rev.speechtotext` sounds good.

PR #16, opened 2020-05-19
AISDK-80: Create ci publishing pipeline for Java SDK

Chanc3:

PR #18, opened 2020-05-20
AISDK-80 - GH Actions build pipeline - dont run integration tests always

dkokotov:


.github/workflows/build_and_test.yml
```diff
@@ -15,25 +15,36 @@ jobs:
...
         java-version: ${{ matrix.java }}
 
-    - name: Build with Maven
-      run: mvn -B package --file pom.xml
+    - name: Build and run tests
+      run: mvn -B package 
+
+    - name: Run integration tests
+      run: mvn verify
+      if: github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release')
```
dkokotov: only run integration tests on develop / master / release branches

src/test/java/ai/rev/speechtotext/integration/EnvHelper.java
```diff
@@ -3,6 +3,6 @@
 public class EnvHelper {
 
     public static String getToken() {
-        return System.getenv("TOKEN");
+        return System.getenv("REVAI_ACCESS_TOKEN");
```
dkokotov, line 6: the less-generic name makes it more convenient to test locally without possible name clashes with other programs for the variable

src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientContentTypeTest.java
```diff
@@ -80,7 +81,7 @@ public void setup() {
     streamContentType.setChannels(channels);
     this.streamContentType = streamContentType;
 
-    streamingClient = new StreamingClient(EnvHelper.getToken());
```
dkokotov, line 83: unit tests dont need a real access token, this way they dont require the environment variable to be set

pom.xml
```diff
@@ -93,7 +93,32 @@
...
                 <version>2.22.1</version>
-            </plugin>
+                <configuration>
+                    <excludes>
+                        <exclude>**/integration/**</exclude>
+                    </excludes>
+                </configuration>
+             </plugin>
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
```
dkokotov, line 118: from what I've read, its preferred to run integration tests with the failsafe plugin

README.md
```diff
@@ -1,6 +1,6 @@
 # Rev.ai Java SDK
 
-[![Build Status](https://travis-ci.org/revdotcom/revai-java-sdk.svg?branch=develop)](https://travis-ci.org/revdotcom/revai-java-sdk)
```
dkokotov, line 3: changed the badge to use GH actions status

.github/workflows/build_and_test.yml
```diff
@@ -15,25 +15,36 @@ jobs:
...
+      uses: actions/setup-java@v1
       with:
         java-version: ${{ matrix.java }}
 
-    - name: Build with Maven
-      run: mvn -B package --file pom.xml
+    - name: Build and run tests
+      run: mvn -B package 
+
+    - name: Run integration tests
```
kbridbur, line 42: How does this decide which environment to run the integration tests against? Or is it always production

dkokotov: i hadnt changed that part, so yes, i think it's always against production.

it looks like for node sdk you run it against production if on a tag, stage if on develop. i think we can add this in a separate PR / issue, just to keep the scope of this effort limited

.github/workflows/build_and_test.yml
```diff
@@ -1,39 +1,79 @@
...
+      if: failure() && env.IS_NON_FEATURE_BRANCH == 'true'
+      env:
+        SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+      uses: voxmedia/github-action-slack-notify-build@v1
+      with:
+        channel: revai-alerts-prod
+        status: FAILED
+        color: danger
+
+  slack_success:
```
dkokotov, line 67: i have the failure notification as a step in the build job, but success notification as a separate job, this way it only gets sent once instead of 3 times (because build job uses matrix strategy to test with multiple java versions).

it does mean failure would likely get sent multiple times because probably most errors would happen on all java versions. we can probably fix it but it is a bit tricky so I am deferring it for a future issue, hopefully build failures are not that frequent.

.github/workflows/build_and_test.yml
kbridbur: Sounds reasonable. I just want to make sure that when I commit a new feature to develop it doesn't fail because we haven't released that feature to the api yet

.github/workflows/build_and_test.yml
```diff
@@ -1,39 +1,79 @@
...
+      run: mkdir staging && cp target/*.jar staging
     
-    - uses: actions/upload-artifact@v1
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v1
       with:
         name: Package
         path: staging
+    
+    - name: Notify slack fail
```
kbridbur, line 57: Can we include the version that failed in this notification?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,34 @@
+# This workflow will build a package using Maven and then publish it to GitHub packages when a release is created
```
kbridbur, line 1: :0 wow this is nice

kbridbur (approved):
lgtm. A few questions. I'd like chance to take a look and give the final shippit since I don't have much experience with maven or java

.github/workflows/build_and_test.yml
dkokotov: yeah it does already, you can see how it looks in the channel - https://rev.slack.com/archives/CFPT8VA2Z/p1590010926249400. it includes the branch name and the link includes the commit

.github/workflows/publish.yml
dkokotov: assuming it works, havent tried this one yet

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+        AWS_DEFAULT_REGION: us-west-2
+        SECRET_NAME: ${{ secrets.MAVEN_PUBLISHING_CREDENTIALS_SECRET_NAME }}
+
+    - name: Release Maven package
+      uses: samuelmeuli/action-maven-publish@v1
+      with:
+        gpg_private_key: ${{ secrets.PGP_PRIVATEKEY }}
+        gpg_passphrase: ${{ secrets.PGP_PASSPHRASE }}
+        nexus_username: ${{ secrets.NEXUS_USERNAME }}
+        nexus_password: ${{ secrets.NEXUS_PASSWORD }}
```
Chanc3: git complaining about empty line at EOF

Chanc3 (approved):
Looks really good. If you merge the pom additions from https://github.com/revdotcom/revai-java-sdk/pull/16 and get the publishing check working. I think we are good to go

.github/workflows/publish.yml
```diff
@@ -0,0 +1,66 @@
...
+        settings-path: ${{ github.workspace }} # location for the settings.xml file
+
+    - name: Build and run tests
+      run: mvn -B package 
+
+    - name: Run the integration tests
+      run: mvn -B package 
+
+    - name: Retrieve Credentials for Publishing Maven package
+      uses: say8425/aws-secrets-manager-actions@v1
```
dkokotov, line 32: one unfortunate thing about this action is that it does not support assuming a role.

I wasted a LOT of time trying to use https://github.com/aws-actions/configure-aws-credentials and then a fork of this action to use an assumed role for credentials, but just could not get it to work.

so for now I attached a policy giving access to the secret directly to the user on whose behalf this is running.

I could spend a bit more time trying to get my fork to work by calling the assume role API directly in the action instead of relying on https://github.com/aws-actions/configure-aws-credentials, OTOH I am not sure its worth it, and the very least i didnt want to be blocked in the short term

PR #19, opened 2020-06-09
AISDK-97: Add custom vocabulary client

Chanc3:


README.md
```diff
@@ -218,6 +218,30 @@ streamingClient.sendAudioData(ByteString);
 
 The streaming connection will close when you call the method `streamingClient.close()` or if you go 15 seconds without sending any audio data.
 
+## Custom Vocabulary (for streaming)
+
+You can now submit any custom vocabularies independently through the new CustomVocabularies client! The main benefit is that users of the SDK can now submit their custom vocabularies for pre-processing and then include these processed custom vocabularies in their streaming jobs.
```
amikofalvy: I would refrain from using adjectives "new" or "old" to describe entities. These terms don't hold up over more than a short time window. 1 year from now, no one will know the difference between "new" or "old" naming.

amikofalvy: Strive to write timeless documentation that someone 10 years from now will be able to read without confusion.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,141 @@
...
+import ai.rev.speechtotext.models.CustomVocabularyInformation;
+import ai.rev.speechtotext.models.asynchronous.RevAiJobOptions;
+import okhttp3.OkHttpClient;
+import retrofit2.Retrofit;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * The CustomVocabularyClient object provides methods to submit custom vocabularies and retrieve
```
amikofalvy: You could make the class description more generic. Leave it to the individual methods to describe their own purpose.

```suggestion
 * The CustomVocabularyClient object provides methods to interact with the Custom Vocabulary Api
```

src/main/java/ai/rev/speechtotext/ApiInterface.java
```diff
@@ -54,4 +55,13 @@
           @QueryMap Map<String, String> query,
           @HeaderMap Map<String, String> contentType
   );
+
+  @POST("vocabularies")
```
amikofalvy: Since the CV API is a different api, can we define the methods in their own interface?

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,141 @@
...
+    if (accessToken == null) {
+      throw new IllegalArgumentException("Access token must be provided");
+    }
+    this.client = ClientHelper.createOkHttpClient(accessToken);
+    Retrofit retrofit = ClientHelper.createRetrofitInstance(client);
+    this.apiInterface = retrofit.create(ApiInterface.class);
+  }
+
+  /**
+   * This methods makes a POST request to the /vocabularies endpoint, starts an asynchronous job to
```
amikofalvy:
```suggestion
   * This methods makes a POST request to the /vocabularies endpoint, starts an asynchronous custom vocabulary job to
```

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,141 @@
...
+   * @param metadata Optional information that can be provided.
+   * @param customVocabularies List of {@link CustomVocabulary} objects.
+   * @param callbackUrl The callback url that Rev.ai will send a POST to when the job has finished.
+   * @return A {@link CustomVocabularyInformation} object.
+   * @throws IOException If the response has a status code > 399.
+   * @throws IllegalArgumentException If the list of custom vocabularies is null.
+   * @see CustomVocabularyInformation
+   */
+  public CustomVocabularyInformation submitCustomVocabularies(
+      String metadata, List<CustomVocabulary> customVocabularies, String callbackUrl)
```
amikofalvy: Are these parameters formatted correctly? It seems strange to line break before the first parameter in order to line wrap.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,141 @@
...
+      throws IOException {
```
amikofalvy, line 54: Should we be throwing the more specific  `IllegalArgumentException`?

src/main/java/ai/rev/speechtotext/ApiInterface.java
```diff
@@ -54,4 +55,13 @@
...
+  Call<CustomVocabularyInformation> submitCustomVocabularies(@Body RevAiJobOptions options);
```
amikofalvy: There is no coupling between job options and cv options. This parameter shouldn't be the same RevAiJobOptions object. There should be a separate RevAiCustomVocabularyOptions definition. Some of the properties will have similar names, but this is fine because they are separate APIs with separate options.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
amikofalvy: The other question I have is about the ordering of the parameters. I would normally order them by importance with the optional parameters coming last: `customVocabularies`, `metadata`, `callbackUrl`.

I think that you have chosen this order so that you can override the method so that both metadata and callbackurl can be optional. I think it is only because the types of theparameters are List<CustomVocabulary>, string, string. The last two parameters would be indistinguishable if you did not have the order that you have written. Is this correct?

I worry that in the effort to create an exhaustive list of overrides we have unintentionally made the ordering of the parameters too confusing. If necessary, I would prefer to have a non-exhastive list of overrides and a more clear parameter ordering. With just a single override method, extensive customization of the parameters can be done by configuring a CustomVocabularyJobOption object with the desired properties and injecting it as a parameter to the function.

src/main/java/ai/rev/speechtotext/models/CustomVocabularyInformation.java
```diff
@@ -0,0 +1,201 @@
...
+import ai.rev.speechtotext.models.asynchronous.RevAiJobStatus;
+import com.google.gson.annotations.SerializedName;
+
+public class CustomVocabularyInformation {
+
+  @SerializedName("id")
+  private String id;
+
+  @SerializedName("status")
+  private RevAiJobStatus status;
```
amikofalvy: I think we need a new object definition for CustomVocab status.

src/main/java/ai/rev/speechtotext/models/CustomVocabularyInformation.java
```diff
@@ -0,0 +1,201 @@
...
+  private String completedOn;
+
+  @SerializedName("metadata")
+  private String metadata;
+
+  @SerializedName("callback_url")
+  private String callbackUrl;
+
+  @SerializedName("failure")
+  private RevAiFailureType failure;
```
amikofalvy: We need cv specific failure type object too.

src/main/java/ai/rev/speechtotext/helpers/ClientHelper.java
```diff
@@ -0,0 +1,29 @@
+package ai.rev.speechtotext.helpers;
+
+import ai.rev.speechtotext.ApiInterceptor;
+import ai.rev.speechtotext.ErrorInterceptor;
+import okhttp3.OkHttpClient;
+import retrofit2.Retrofit;
+import retrofit2.converter.gson.GsonConverterFactory;
+import retrofit2.converter.scalars.ScalarsConverterFactory;
+
+public class ClientHelper {
```
amikofalvy, line 10: Can this be used by the async api client as well?

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
```diff
@@ -0,0 +1,238 @@
...
+    mockInterceptor.setSampleResponse(gson.toJson(mockCustomVocabularyInfo));
+
+    List<String> phrases = Arrays.asList("test", "custom", "vocabulary");
+    CustomVocabulary customVocabulary = new CustomVocabulary(phrases);
+
+    CustomVocabularyInformation customVocabularyInformation =
+        sut.submitCustomVocabularies(
+            testName.getMethodName(), Collections.singletonList(customVocabulary), CALLBACK_URL);
+
+    assertRequest(VOCABULARY_URL, "POST");
```
amikofalvy: I think you're missing an assertion on the request body. It seems we're only checking for url and apiMethod

amikofalvy (changes requested):
I left a few suggestions inline.

README.md
Chanc3: Funny. I pulled this from our other SDKs. I will edit and file some tickets to update the python and node readme's

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
Chanc3: Your reasoning as to the ordering is correct. I too was feeling that we should have an options object but in order to follow the other SDK implementations it requires this parameter ordering. Are we ok with deviating here?

src/main/java/ai/rev/speechtotext/helpers/ClientHelper.java
Chanc3: Yes! That was the intention. Good catch.

README.md
```diff
@@ -218,6 +218,30 @@ streamingClient.sendAudioData(ByteString);
...
+
+Below you can see an example of how to create and submit as well as check on the status and other associated information of your submitted custom vocabulary!
+
+```
+// Initialize your client with your rev.ai access token
+String accessToken = "Your Access Token";
+CustomVocabulariesClient customVocabulariesClient = new CustomVocabulariesClient(accessToken);
+
+// Construct a CustomVocabulary object using your desired phrases
```
jennywong2129, line 232: Should you mention the limits on how many phrases you can have in a CV/how many CVs can be used for a job?

src/main/java/ai/rev/speechtotext/models/CustomVocabularyInformation.java
```diff
@@ -0,0 +1,201 @@
...
+   * ISO-8601 UTC form.
+   *
+   * @return A String that contains the date and time the custom vocabulary was completed on.
+   */
+  public String getCompletedOn() {
+    return completedOn;
+  }
+
+  /**
+   * Set the date and time the custom vocabulary was completed on.
```
jennywong2129: Sets*

README.md
Chanc3: I think we are not mentioning the things like limits on purpose to prevent having to update the documentation in too many places.

src/main/java/ai/rev/speechtotext/AsyncApiInterface.java
```diff
@@ -16,7 +25,7 @@
  * The ApiInterface is a type-safe Retrofit interface that presents all the endpoints that are made
  * to communicate with the Rev.AI API.
  */
-public interface ApiInterface {
+public interface AsyncApiInterface {
```
amikofalvy: Since this is public access changing the name of this interface would be a breaking change, right?

src/main/java/ai/rev/speechtotext/CustomVocabularyApiInterface.java
```diff
@@ -0,0 +1,23 @@
...
+import ai.rev.speechtotext.models.vocabulary.CustomVocabularyOptions;
+import retrofit2.Call;
+import retrofit2.http.Body;
+import retrofit2.http.GET;
+import retrofit2.http.POST;
+import retrofit2.http.Path;
+
+import java.util.List;
+
+public interface CustomVocabularyApiInterface {
```
amikofalvy, line 17: I think this needs documentation like the proceeding ApiInterface: https://github.com/revdotcom/revai-java-sdk/pull/19/files#diff-bc6d97b48980ad0bde766336dd7e5d6bR25

src/main/java/ai/rev/speechtotext/clients/ApiClient.java
```diff
@@ -1,18 +1,18 @@
-package ai.rev.speechtotext;
+package ai.rev.speechtotext.clients;
```
amikofalvy: I think changing the package is another breaking change.

src/main/java/ai/rev/speechtotext/clients/ApiClient.java
```diff
@@ -31,7 +31,7 @@
   private OkHttpClient client;
 
   /** Interface that ApiClient methods use to make requests */
-  public ApiInterface apiInterface;
+  public AsyncApiInterface asyncApiInterface;
```
amikofalvy: renaming any public member is a breaking change, right?

amikofalvy: Maybe these should be private?

src/main/java/ai/rev/speechtotext/AsyncApiInterface.java
Chanc3: yes you are right. I'll have to revert these organizational changes and renaming.

src/main/java/ai/rev/speechtotext/clients/ApiClient.java
Chanc3: Unfortunately because the unit tests need the interfaces to create mock retrofit objects we can't make them private.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Constructs the custom vocabulary client used to create custom vocabularies for streaming. The
+   * user access token can be generated on the website at <a
+   * href="https://www.rev.ai/access_token">https://www.rev.ai/access_token</a>.
+   *
+   * @param accessToken Rev.ai authorization token associate with the account.
+   */
+  public CustomVocabulariesClient(String accessToken) {
+    if (accessToken == null) {
```
KostasRev: Given that we are doing this check for the customer, might as well check that the `accessToken` is not empty.

src/main/java/ai/rev/speechtotext/ApiInterface.java
```diff
@@ -46,12 +55,11 @@
   @Multipart
   @POST("jobs")
   Call<RevAiJob> submitJobLocalFile(
-    @Part MultipartBody.Part file, @Part("options") RevAiJobOptions options);
+      @Part MultipartBody.Part file, @Part("options") RevAiJobOptions options);
```
KostasRev, line 58: Curious why you indented this 4 spaces and not 2. Is this according to the style convention? We follow Google style convention right?

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    this.client = ClientHelper.createOkHttpClient(accessToken);
+    Retrofit retrofit = ClientHelper.createRetrofitInstance(client);
+    this.customVocabularyApiInterface = retrofit.create(CustomVocabularyApiInterface.class);
+  }
+
+  /**
+   * This methods makes a POST request to the /vocabularies endpoint, starts an asynchronous custom
+   * vocabulary job to process the vocabularies and returns a {@link CustomVocabularyInformation}
+   * object.
```
KostasRev:
```suggestion
   * This method makes a POST request to the /vocabularies endpoint and returns a {@link CustomVocabularyInformation}
   * object that provides details about the custom vocabulary submission and its progress.
```

I think it might be unnecessary to describe the "asynchronous custom vocabulary job to process the vocabularies" part which happens on the server side.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,89 @@
...
+   * @param options A object that contains the custom vocabularies as well as optional metadata and
+   *     callback url
+   * @return A {@link CustomVocabularyInformation} object.
+   * @throws IOException If the response has a status code > 399.
+   * @throws IllegalArgumentException If the list of custom vocabularies is null.
+   * @see CustomVocabularyInformation
+   */
+  public CustomVocabularyInformation submitCustomVocabularies(CustomVocabularyOptions options)
+      throws IOException {
+    if (options.getCustomVocabularies() == null) {
```
KostasRev: Similarly here, we can check if the list is empty. Update the docstring as well if you make that change.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
KostasRev: I think it is preferable to use the options object rather than mirror the other SDKs; it is better code practice. I think the reason we did not go down that path in the other SDKs is that we would have introduced a breaking change.

src/main/java/ai/rev/speechtotext/models/vocabulary/CustomVocabularyOptions.java
```diff
@@ -0,0 +1,93 @@
+package ai.rev.speechtotext.models.vocabulary;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.List;
+
+public class CustomVocabularyOptions {
```
KostasRev: I think it would be better to name this class `SubmitCustomVocabularyOptions`

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
```diff
@@ -0,0 +1,283 @@
...
+            .baseUrl("https://api.rev.ai/revspeech/v1/")
+            .addConverterFactory(GsonConverterFactory.create())
+            .client(mockOkHttpClient)
+            .build();
+    sut.customVocabularyApiInterface = mockRetrofit.create(CustomVocabularyApiInterface.class);
+  }
+
+  @Test
+  public void
+      submitCustomVocabulary_WhenVocabularyCallbackMetadataAreSpecified_ReturnsACustomVocabularyInformation()
```
KostasRev: Hmm, is it conventional to put the method name on separate line from its access modifier and return type? It looks weird. Test names are expected to be long so I wouldn't worry about line length minimization if convention is not at play here.

KostasRev:
```suggestion
  public void
      submitCustomVocabulary_WhenVocabularyCallbackAndMetadataAreSpecified_ReturnsACustomVocabularyInformation()
```

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
```diff
@@ -0,0 +1,283 @@
...
+
+    mockInterceptor.setSampleResponse(gson.toJson(mockCustomVocabularyInfo));
+
+    List<String> phrases = Arrays.asList("test", "custom", "vocabulary");
+    CustomVocabulary customVocabulary = new CustomVocabulary(phrases);
+
+    CustomVocabularyOptions options = new CustomVocabularyOptions();
+    options.setCustomVocabularies(Collections.singletonList(customVocabulary));
+    options.setMetadata(testName.getMethodName());
+    options.setCallbackUrl(CALLBACK_URL);
```
KostasRev: It would be better to extract the large `setup` blocks of code into private helper methods if possible. With that approach you can also have all the private helper method calls be packed at the top of the method together which helps keep the test easy to read with a clear visual separation of the "arrange/act/assert" sections.

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
KostasRev: Another note is that you can drop the "When" in the condition. Rev convention in tests is "method_condition_result" so I think the "when" is always understood by implication.

src/main/java/ai/rev/speechtotext/clients/ApiClient.java
KostasRev: Can they be protected?

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -0,0 +1,89 @@
...
+   * CustomVocabularyInformation} object.
+   *
+   * @param id The Id of the custom vocabulary to return an object for.
+   * @return A {@link CustomVocabularyInformation} object.
+   * @throws IOException If the response has a status code > 399.
+   * @throws IllegalArgumentException If the custom vocabulary Id is null.
+   * @see CustomVocabularyInformation
+   */
+  public CustomVocabularyInformation getCustomVocabularyInformation(String id) throws IOException {
+    if (id == null) {
```
KostasRev: Can check if string is empty here as well.

KostasRev (changes requested):
Lookin good, I made some suggestions. Please manually test the client as well after all the changes.

src/main/java/ai/rev/speechtotext/clients/ApiClient.java
Chanc3: Unfortunately you cannot declare the class protected. But even if it were, that would make this class accessible only within the same package which would prevent the unit tests from using it :(

src/main/java/ai/rev/speechtotext/ApiInterface.java
Chanc3: Yup this is the [google style convention](https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent).

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
Chanc3: This line wrapping here follows the google format. The only exceptions made are listed out [here](https://google.github.io/styleguide/javaguide.html#s4.4-column-limit).

I can drop the when and will do so for all our other tests as well.

src/main/java/ai/rev/speechtotext/models/vocabulary/CustomVocabularyOptions.java
Chanc3: I'm torn here. That name sounds like you are submitting options. Since the `CustomVocabulary` is required the `Options` part sounds misleading.

What about something like `CustomVocabularyContents`? What we are submitting is the contents of the vocabulary...

KostasRev: Haha I see your point. I was trying alternative names because I think `CustomVocabularyOptions` sounds like it is just the `Options` part rather than also the `CustomVocabulary` part (ie the phrases) which are core to the submission. 

What if we call it `CustomVocabularySubmission`? That makes it clear that these are not just options.

src/main/java/ai/rev/speechtotext/ApiClient.java
```diff
@@ -37,10 +37,10 @@
    * href="https://www.rev.ai/access_token">https://www.rev.ai/access_token</a>.
    *
    * @param accessToken Rev.ai authorization token associate with the account.
-   * @throws IllegalArgumentException If the access token is null.
+   * @throws IllegalArgumentException If the access token is null or empty.
    */
   public ApiClient(String accessToken) {
-    if (accessToken == null) {
+    if (accessToken == null && accessToken.isEmpty()) {
```
KostasRev: All of these emptiness checks should be `||` rather than `&&`. The purpose is to verify that the accessToken given is nonempty or null.

Please change this and all the other `&& object.isEmpty()` checks.

src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
```diff
@@ -82,62 +77,38 @@ public void setup() {
...
+      submitCustomVocabulary_VocabularyAndCallbackAreSpecified_ReturnsACustomVocabularyInformation()
           throws IOException {
-    CustomVocabularyInformation mockCustomVocabularyInfo = new CustomVocabularyInformation();
-    mockCustomVocabularyInfo.setCallbackUrl(CALLBACK_URL);
-    mockCustomVocabularyInfo.setCreatedOn(DATE);
-    mockCustomVocabularyInfo.setId(ID);
-    mockCustomVocabularyInfo.setStatus(CustomVocabularyStatus.IN_PROGRESS);
-
+    CustomVocabularyInformation mockCustomVocabularyInfo =
+        createCustomVocabularyInfo(DATE, ID, IN_PROGRESS, null, CALLBACK_URL);
```
KostasRev, line 105: Nice.

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Test
+  public void
+      submitCustomVocabulary_MetadataAndVocabAreIncluded_ReturnCustomVocabularyInformationInProgress() {
+    CustomVocabularyInformation customVocabularyInformation = submitCustomVocabulary();
+
+    assertThat(customVocabularyInformation.getStatus())
+        .as("Custom vocabulary status")
+        .isEqualTo(CustomVocabularyStatus.IN_PROGRESS);
+  }
```
KostasRev: Why is callback not included in this test? I would add it here.

Also I would separate the `customVocabularyInformation.getStatus()` out of the assert statement, please do the same other tests that do this as well. It is useful to have 3 visual blocks in tests:

```
     // Arrange
     CustomVocabularyInformation customVocabularyInformation = submitCustomVocabulary(); 

     // Act
     status = customVocabularyInformation.getStatus()

    // Assert ...
```

KostasRev (changes requested):
One last set of changes and I think we can ship afterwards! 

Most notably the `&&` to `||` for any place that is checking for emptiness.

src/main/java/ai/rev/speechtotext/ApiClient.java
Chanc3: oh man good catch. I don't know what I was thinking.

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
Chanc3: The callback wasn't included because the integration test really sends a request and I don't have a valid callback url to pass in. This same logic applied to the `SubmitJobTest` integration tests. Should we just add something like `https://test.test.test`?

Are you sure about the blocks? I feel like this is a style preference and adds some unnecessary assignment. I feel like because the everything is named well it's easy to understand and reads well. `Assert that custom vocabulary information status is equal to the custom vocabulary status in_progress`, no?

Also blocks like:
```
for (CustomVocabularyInformation customVocabulary : customVocabularies) {
      assertThat(customVocabulary.getId()).as("Vocabulary Id").isNotNull();
      assertThat(customVocabulary.getCreatedOn()).as("Created on").isNotNull();
      assertThat(customVocabulary.getStatus()).as("Status").isNotNull();
      if (customVocabulary.getStatus() == CustomVocabularyStatus.FAILED) {
        assertThat(customVocabulary.getFailure()).as("Failure").isNotNull();
        assertThat(customVocabulary.getFailureDetail()).as("Failure detail").isNotNull();
      } else if (customVocabulary.getStatus() == CustomVocabularyStatus.COMPLETE) {
        assertThat(customVocabulary.getCompletedOn()).as("Completed on").isNotNull();
      }
    }
```

Will become
```
for (CustomVocabularyInformation customVocabulary : customVocabularies) {
    String id = customVocabulary.getId();
    String createdOn = customVocabulary.getCreatedOn();
    String completedOn = customVocabulary.getCompletedOn();
    CustomVocabularyStatus status = customVocabulary.getStatus();
    CustomVocabularyFailureType failure = customVocabulary.getFailure();
    String failureDetail = customVocabulary.getFailureDetail();

    assertThat(id).as("Vocabulary Id").isNotNull();
    assertThat(createdOn).as("Created on").isNotNull();
    assertThat(status).as("Status").isNotNull();
    if (status == CustomVocabularyStatus.FAILED) {
      assertThat(failure).as("Failure").isNotNull();
      assertThat(failureDetail).as("Failure detail").isNotNull();
    } else if (status == CustomVocabularyStatus.COMPLETE) {
      assertThat(completedOn).as("Completed on").isNotNull();
    }
  }
```

In these two cases I'm not really seeing a big difference but if you think this improves readability I'll make the changes everywhere. 

Will we also need to do it in other SDKs? What about tests in node like https://github.com/revdotcom/revai-node-sdk/blob/develop/test/integration/test/custom_vocabularies.test.js

Chanc3: This request was clarified over a zoom and changes have been implemented.

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -41,20 +41,21 @@ public CustomVocabulariesClient(String accessToken) {
    * CustomVocabularyInformation} object that provides details about the custom vocabulary
    * submission and its progress.
    *
-   * @param options A object that contains the custom vocabularies as well as optional metadata and
-   *     callback url
+   * @param submission A object that contains the custom vocabularies as well as optional metadata
+   *     and callback url
```
KostasRev: Couple of minor things - "A object" -> "An object"

Also, I would not mention the optional fields in the docstring, this list would keep growing for one thing everytime we add a new optional parameter. I think it would suffice to say something along the lines of "An object that contains the custom vocabularies as well as some optional parameters"

Also would it help to add a {@link CustomVocabularySubmission}?

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
```diff
@@ -27,7 +27,10 @@
   @Test
   public void
       submitCustomVocabulary_MetadataAndVocabAreIncluded_ReturnCustomVocabularyInformationInProgress() {
```
KostasRev:
```suggestion
      submitCustomVocabulary_VocabAndAllOptionsAreIncluded_ReturnCustomVocabularyInformationInProgress() {
```

Need to rename the test to indicate that all options are being specified, feel free to use an alternate phrasing to my suggestion.

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
```diff
@@ -36,67 +39,77 @@
...
     }
 
     assertThat(customVocabularies.size()).as("List size").isGreaterThan(1);
     for (CustomVocabularyInformation customVocabulary : customVocabularies) {
+      CustomVocabularyStatus status = customVocabulary.getStatus();
       assertThat(customVocabulary.getId()).as("Vocabulary Id").isNotNull();
       assertThat(customVocabulary.getCreatedOn()).as("Created on").isNotNull();
-      assertThat(customVocabulary.getStatus()).as("Status").isNotNull();
-      if (customVocabulary.getStatus() == CustomVocabularyStatus.FAILED) {
+      assertThat(status).as("Status").isNotNull();
```
KostasRev, line 88: Minor - is this not null check needed given that the status is checked in the next line?

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
```diff
@@ -36,67 +39,77 @@
...
+      if (status == CustomVocabularyStatus.FAILED) {
         assertThat(customVocabulary.getFailure()).as("Failure").isNotNull();
         assertThat(customVocabulary.getFailureDetail()).as("Failure detail").isNotNull();
```
KostasRev: I wonder if there is utility to check that the failure is a member of `CustomVocabularyFailureType` 🤔

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
```diff
@@ -36,67 +39,77 @@
...
       customVocabularies = customVocabularyClient.getListOfCustomVocabularyInformation();
     } catch (IOException e) {
-      e.printStackTrace();
+      throw new RuntimeException(
+          "Error occurred getting the list of vocabularies: " + e.getMessage());
     }
 
     assertThat(customVocabularies.size()).as("List size").isGreaterThan(1);
     for (CustomVocabularyInformation customVocabulary : customVocabularies) {
+      CustomVocabularyStatus status = customVocabulary.getStatus();
```
KostasRev, line 85: Good DRYing 😎

src/test/java/ai/rev/speechtotext/unit/AssertHelper.java
```diff
@@ -0,0 +1,36 @@
...
+      throw new RuntimeException(e.getMessage());
+    }
+
+    T requestBodyObject = gson.fromJson(buffer.readUtf8(), expectedClass);
+
+    String requestBodyString = gson.toJson(requestBodyObject);
+    String expectedBodyString = gson.toJson(expectedObject);
+    assertThat(requestBodyString).as("Request body").isEqualTo(expectedBodyString);
+  }
+}
```
KostasRev, line 36: Very nice!

KostasRev (approved):
Looks good to me, and the tests _especially_ so!

But I left some minor comments, please address those, and make sure you've tested all clients that were altered and that they behave as expected and :shipit:

src/test/java/ai/rev/speechtotext/integration/CustomVocabulariesTest.java
Chanc3: I felt it was still needed because is should not be null and the next two if statements only check if it is failed or completed. This will ensure that it is always set to something.

PR #21, opened 2020-06-18
AISDK-107: Add examples to the java SDK

Chanc3:
This PR also:
* Fixes the flakiness of the custom vocab delete test
* Added the filter profanity option to async requests

### DIFF
--- a/examples/AsyncTranscribeMediaUrl.java
+++ b/examples/AsyncTranscribeMediaUrl.java
@@ -36,7 +36,8 @@ public class AsyncTranscribeMediaUrl {
 
     RevAiJob submittedJob;
 
-    String mediaUrl = "https://your.media.url";
+    String mediaUrl =
+        "https://support.rev.com/hc/en-us/article_attachments/200043975/FTC_Sample_1_-_Single.mp3";
 
     try {
       // Submit the local file and transcription options
@@ -49,12 +50,7 @@ public class AsyncTranscribeMediaUrl {
     System.out.println("Job Status: " + submittedJob.getJobStatus());
     System.out.println("Created On: " + submittedJob.getCreatedOn());
 
-    /*
-     * Waits 5 seconds between each status check to see if job is complete.
-     * NOTE: polling for job status is not recommended in a non-testing environment.
-     * Use the callback_url supplied in the RevAiJobOptions object to receive the
-     * response asynchronously on job completion
-     */
+    // Waits 5 seconds between each status check to see if job is complete
     boolean isJobComplete = false;
     while (!isJobComplete) {
       RevAiJob retrievedJob;
--- a/examples/AsyncTranscribeMediaUrl.java
+++ b/examples/AsyncTranscribeMediaUrl.java
@@ -1,0 +1,102 @@
+package ai.rev.speechtotext;
+
+import ai.rev.speechtotext.models.asynchronous.RevAiCaptionType;
+import ai.rev.speechtotext.models.asynchronous.RevAiJob;
+import ai.rev.speechtotext.models.asynchronous.RevAiJobOptions;
+import ai.rev.speechtotext.models.asynchronous.RevAiJobStatus;
+import ai.rev.speechtotext.models.asynchronous.RevAiTranscript;
+import ai.rev.speechtotext.models.vocabulary.CustomVocabulary;
+
+import java.io.IOException;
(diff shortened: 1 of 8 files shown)


Chanc3:
A few questions I have are:
- Do we need to include audio files to use in the examples? (Python does not but Node does)
- Should I include examples for GET /jobs and GET /account? (These two don't really seem necessary)

examples/AsyncTranscribeMediaUrl.java
```diff
@@ -0,0 +1,102 @@
...
+    revAiJobOptions.setMetadata("My first submission");
+    revAiJobOptions.setCallbackUrl("https://example.com");
+    revAiJobOptions.setSkipPunctuation(false);
+    revAiJobOptions.setSkipDiarization(false);
+    revAiJobOptions.setFilterProfanity(true);
+    revAiJobOptions.setSpeakerChannelsCount(null);
+
+    RevAiJob submittedJob;
+
+    String mediaUrl = "https://your.media.url";
```
amikofalvy: You can use the example media url we use in other places for rev.ai. It contains a real media file.
`https://support.rev.com/hc/en-us/article_attachments/200043975/FTC_Sample_1_-_Single.mp3`

examples/AsyncTranscribeMediaUrl.java
```diff
@@ -0,0 +1,102 @@
...
+      throw new RuntimeException("Failed to submit url [" + mediaUrl + "] " + e.getMessage());
+    }
+    String jobId = submittedJob.getJobId();
+    System.out.println("Job Id: " + jobId);
+    System.out.println("Job Status: " + submittedJob.getJobStatus());
+    System.out.println("Created On: " + submittedJob.getCreatedOn());
+
+    /*
+     * Waits 5 seconds between each status check to see if job is complete.
+     * NOTE: polling for job status is not recommended in a non-testing environment.
```
amikofalvy: We don't disallow or recommend against polling for all customers. I would remove this line and description of best practices for when to use webhooks.

examples/CustomVocabulariesExample.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    String customVocabularyId = submittedCustomVocabularyInfo.getId();
+    System.out.println("Vocabulary Id: " + customVocabularyId);
+    System.out.println("Vocabulary Status: " + submittedCustomVocabularyInfo.getStatus());
+    System.out.println("Created On: " + submittedCustomVocabularyInfo.getCreatedOn());
+
+    /*
+     * Waits 5 seconds between each status check to see if vocabulary has been processed.
+     * NOTE: polling for status is not recommended in a non-testing environment.
```
amikofalvy: ditto

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -44,6 +44,10 @@
   @SerializedName("metadata")
   private String metadata;
 
+  /** Optional parameter to filter profanity in the transcript */
+  @SerializedName("filter_profanity")
```
amikofalvy, line 48: I think it makes sense to add these in a separate PR.
Landing logically separate improvements as separate commits allows us to easily rollback changes with git cli and provides a better git history with clear commit messages.

amikofalvy: for example: we could also add remove_disfluencies option, but I think it would be best if that happened in a separate commit

amikofalvy: But if filter_profanity is already implemented in this pr then no reason to delay by taking it out. we'll just keep it in mind for the future.

amikofalvy (approved):
> Do we need to include audio files to use in the examples? (Python does not but Node does)

I think it is fine without files for local submission. I did suggest that you change the media_urls to a valid url.

> Should I include examples for GET /jobs and GET /account? (These two don't really seem necessary)

I don't think these two endpoints require examples.

Once all feedback is addressed, and I'm giving the PR a :shipit: 

Please have a third party verify the examples (Raabya). It is critical that these examples work and are useable by someone other than the author. QA can help us verify the examples work as they expect.

PR #22, opened 2020-06-30
AISDK-115: Add remove_disfluencies to Java sdk

timitijani:
* Added remove_disfluencies option to Async and Streaming clients
* Updated examples to use the new option
* Updated tests (unit + integration)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,7 +37,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.11</version>
+            <version>4.12</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>ai.rev.speechtotext</groupId>
     <artifactId>revai-java-sdk-speechtotext</artifactId>
-    <version>1.0.0</version>
+    <version>1.1.0</version>
     <name>Rev.ai speech to text SDK for Java</name>
     <description>Java SDK for rev.ai speech to text API</description>
     <url>https://www.rev.ai/docs</url>
@@ -22,7 +22,7 @@
                 <role>developer</role>
             </roles>
         </developer>
-    </developers>    
+    </developers>
     <scm>
         <connection>scm:git:https://github.com/revdotcom/revai-java-sdk.git</connection>
         <developerConnection>scm:git:ssh://github.com/revdotcom/revai-java-sdk.git</developerConnection>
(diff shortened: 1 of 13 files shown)


timitijani:
Bloated by running `Refactor Code` which was mentioned in README

src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientOptionalParametersTest.java
```diff
@@ -27,103 +27,115 @@
...
+                        {METADATA, VOCAB_ID, true, null},
+                        {METADATA, VOCAB_ID, false, null},
+                        {METADATA, VOCAB_ID, null, null},
+                        {METADATA, null, true, true},
+                        {null, VOCAB_ID, true, false},
+                        {null, VOCAB_ID, false, null},
+                        {null, VOCAB_ID, null, true},
+                        {null, null, true, true},
+                        {null, null, false, false},
+                        {null, null, null, null}
```
KostasRev: Please check if there's another way to specify these option sets, this matrix is not sustainable. If possible we should change it starting from this PR.

KostasRev (changes requested):
In addition to the docs change we mentioned, please see the parametrization comment, but otherwise looks good!

Make sure to locally test against API after making all the changes.

src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientOptionalParametersTest.java
amikofalvy: ☝️ I agree

timitijani: I changed it to `@Theory` and it tests all the cases.

amikofalvy (approved):
Please verify that a broken test case using theories fails the test and shows up as a failure in the test suite. Afterward, :shipit:

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>ai.rev.speechtotext</groupId>
     <artifactId>revai-java-sdk-speechtotext</artifactId>
-    <version>1.0.0</version>
+    <version>1.1.0</version>
```
KostasRev, line 7: Please double check with @Chanc3 that this is the proper way to bump the version, but seems reasonable to me.

src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientOptionalParametersTest.java
```diff
@@ -10,120 +10,125 @@
...
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@RunWith(Parameterized.class)
+@RunWith(Theories.class)
```
KostasRev, line 25: Fantastic, much better now!

KostasRev (approved):
Looking much better with the theory now! Nice work, please address Andrew's suggestion and 🚢

pom.xml
timitijani: Talked to @Chanc3 , this is the correct way to bump the version.

PR #24, opened 2020-07-22
AISDK-120: fix fragile test

Chanc3:


src/test/java/ai/rev/speechtotext/integration/GetCaptionsTest.java
```diff
@@ -45,18 +48,31 @@ public void GetCaptions_TypeIsSpecifiedAndJobIdIsValid_ReturnsCaptionStreamInVtt
     assertThat(captions).as("Caption content").contains(VTT_CONTAINS);
   }
 
-  private String getTranscribedJob() throws IOException {
-    List<RevAiJob> jobs = apiClient.getListOfJobs();
-    String transcribedJobId = null;
-    for (RevAiJob job : jobs) {
-      if (job.getJobStatus().equals(RevAiJobStatus.TRANSCRIBED)) {
-        transcribedJobId = job.getJobId();
+  private void pollForJobCompletionForFiveMinutes(String jobId) {
```
amikofalvy: Because you have named this function to include time I would use the elapsed time instead of pollingAttempts * sleepDuration to configure the timeout. It makes the function less susceptible to error when a future developer changes.

You could also change the name of the function to match the implementation.

Additionally, I would throw some type of timeout exception if the operation times out.

amikofalvy (approved):
One comment about clarifying the name of the helper function then :shipit:

amikofalvy (approved):

PR #28, opened 2020-12-30
AISDK-138: Add Language parameter to the Java SDK

lrgottlieb:


src/test/java/ai/rev/speechtotext/integration/SubmitJobTest.java
```diff
@@ -148,6 +148,7 @@ private RevAiJobOptions getJobOptions() {
     revAiJobOptions.setCallbackUrl("https://example.com");
     revAiJobOptions.setSpeakerChannelsCount(null);
     revAiJobOptions.setDeleteAfterSeconds(0);
+    revAiJobOptions.setLanguage("en");
```
Chanc3, line 151: Looks like we are only testing english transcription. @amikofalvy I noticed that we aren't testing the other languages in the other SDKs either. What are your thoughts around including integration tests for other languages?

Are we relying on the [revai api tests](https://github.com/revdotcom/revai-tests/blob/master/src/test/java/com/revai/api/submission/RevAiForeignLanguagesSubmissionValidation.java) to catch those?

amikofalvy: The SDK integration tests are testing the SDK's ability to interact with the API. As you point out, the integration test suite for the API covers the testing for the other languages.

amikofalvy: I don't see a reason to have duplicate coverage which performs real transcription for foreign languages. It will not provide additional benefit and it will cost additional dollars.

amikofalvy (changes requested):
You need to increase the package version. The current published package is 1.2.0. I think we should bump the minor version to 1.3.0.

examples/AsyncTranscribeLocalMediaFile.java
```diff
@@ -35,6 +35,7 @@ public static void main(String[] args) {
     revAiJobOptions.setRemoveDisfluencies(true);
     revAiJobOptions.setSpeakerChannelsCount(null);
     revAiJobOptions.setDeleteAfterSeconds(2592000); // 30 days in seconds
+    revAiJobOptions.setLanguage("en");
```
amikofalvy, line 38: Did you run these examples manually to ensure they still work?

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -40,22 +40,36 @@
...
+  /**
+   * Optional parameter to remove disfluencies (ums, ahs) in the transcript
+   */
   @SerializedName("remove_disfluencies")
   private Boolean removeDisfluencies;
 
-  /** Optional number of seconds after job completion when job is auto-deleted */
+  /**
+   * Optional number of seconds after job completion when job is auto-deleted
+   */
```
amikofalvy, line 63: These formatting changes are not really required.

amikofalvy (approved):
Please manually test the examples with non english language value to ensure your changes have the desired effect, then :shipit:

PR #29, opened 2021-08-05
AISDK-158: Change language description in Java in-line doc

kshiflett88:


examples/AsyncTranscribeLocalMediaFile.java
```diff
@@ -35,7 +35,8 @@ public static void main(String[] args) {
     revAiJobOptions.setRemoveDisfluencies(true);
     revAiJobOptions.setSpeakerChannelsCount(null);
     revAiJobOptions.setDeleteAfterSeconds(2592000); // 30 days in seconds
-    revAiJobOptions.setLanguage("en");
+    revAiJobOptions.setLanguage("en"); // Specify language using the one of the supported ISO 639-1 (2-letter) or
+                                       // ISO 639-3 (3-letter) language codes as defined in the API Reference
```
k-weng: I don’t think these are the places we need to add the inline comments - looking up `language` in the repo, there’s a `RevAiJobOptions` that contain the parameter for `language` and I think that is where we should be updating the documentation.

https://github.com/revdotcom/revai-java-sdk/blob/a87eeea4f351f6a9871c4a9acdc2153cb21cc2ca/src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java#L67-L71

k-weng (changes requested):
Should update documentation in `RevAiJobOptions`, not in the examples

examples/AsyncTranscribeLocalMediaFile.java
```diff
@@ -35,7 +35,7 @@ public static void main(String[] args) {
     revAiJobOptions.setRemoveDisfluencies(true);
     revAiJobOptions.setSpeakerChannelsCount(null);
     revAiJobOptions.setDeleteAfterSeconds(2592000); // 30 days in seconds
-    revAiJobOptions.setLanguage("en");
+    revAiJobOptions.setLanguage("en");
```
k-weng: Remove whitespace
```suggestion
    revAiJobOptions.setLanguage("en");
```

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -65,7 +65,7 @@
   private Integer deleteAfterSeconds;
 
   /**
-   * Optional language parameter using ISO 639-1 2-letter language code
+   * Optional language parameter using ISO 639-1 2-letter or ISO 639-3 (3-letter) language code
```
k-weng: Since it's a documentation comment, we can add a bit more details
```suggestion
   * Optional language parameter using a supported ISO 639-1 2-letter or ISO 639-3 (3-letter) language code
   * as defined in the API reference
```

k-weng (approved):
Two comments to address then can merge

PR #31, opened 2022-01-28
AISDK-156: Add support for human transcription

k-weng:


src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJob.java
```diff
@@ -145,7 +183,7 @@ public Double getDurationSeconds() {
   /**
    * Sets the audio duration.
    *
-   * @param durationSeconds An Integer value to set as audio duration.
+   * @param durationSeconds An Double value to set as audio duration.
```
amikofalvy:
```suggestion
   * @param durationSeconds A Double value to set as audio duration.
```

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJob.java
```diff
@@ -145,7 +183,7 @@ public Double getDurationSeconds() {
   /**
    * Sets the audio duration.
    *
-   * @param durationSeconds An Integer value to set as audio duration.
+   * @param durationSeconds A Double value to set as audio duration.
```
amikofalvy, line 186: I don't think it is necessary to document type in the JavaDoc.

amikofalvy: We can avoid this type of change if we don't replicate the typing in the javadoc.

k-weng: Was following the established patterns that we had for our other params

amikofalvy (approved):

README.md
```diff
@@ -69,6 +69,21 @@ try {
...
 
+You can also submit a job to be handled by a human transcriber using our [Human Transcription](https://www.rev.ai/docs#section/Human-Transcription-(Labs)) option.
+```
+String urlLinkToFile = "https://www.rev.ai/FTC_Sample_1.mp3";
+RevAiJobOptions options = new RevAiJobOptions();
+
+// set to perform human transcription
+options.setTranscriber("human");
+
+// optional job options
```
kbridbur, line 80: Should segments to transcribe also be included here? The segment from 0 to 2 minutes should work with that file

kbridbur: Potentially test mode as well

kbridbur (approved):
Two small questions. Looks good otherwise

README.md
k-weng: Done

PR #37, opened 2022-03-26
AISDK-182: Add broken link checker to github actions

amikofalvy:
After implementing the previous ticket AISDK-179 (#36), I decided to explore link checking options to reduce verification time and prevent broken links from occurring in future deploys. You can review the Github Actions runs for this new action here: https://github.com/revdotcom/revai-java-sdk/actions/workflows/links-fail-fast.yml.

If we like this approach then we can implement it for our other repositories as well.


src/test/java/ai/rev/speechtotext/unit/RevAiCustomVocabularyTest.java
```diff
@@ -35,7 +35,7 @@
 
   private final MediaType MEDIA_TYPE = MediaType.get("application/json; charset=utf-8");
   private final String VOCABULARY_URL = "https://api.rev.ai/revspeech/v1/vocabularies";
-  private final String CALLBACK_URL = "https://test.callback.url";
+  private final String CALLBACK_URL = "https://www.example.com";
```
amikofalvy, line 38: It's generally best practices to use domain name reserved such examples.

k-weng (approved):

PR #42, opened 2022-04-18
AISDK-190 Update sdks to support notification_config and source_config

aaron-wilson-rev:


README.md
```diff
@@ -144,6 +144,7 @@ apiClient.deleteJob(revAiJob.getJobId());
 
 Once your file is transcribed, you can get your transcript in a few different forms:
 
+
```
amikofalvy: minor - extra space. please remove.

amikofalvy (changes requested):
Let's talk on Zoom about making breaking changes to the SDK.

aaron-wilson-rev:
~~Noticed while testing the examples the CV doesn't currently support notification_config. If we want that we'll need to update that on the api.~~
Added to cv in another ticket

examples/AsyncTranscribeMediaUrl.java
```diff
@@ -24,11 +27,26 @@ public static void main(String[] args) {
...
+
+    // Set up notification configuration parameters
+    String callbackUrl = "https://example.com";
+    // Authorization header is optional; use it if needed to access the callback url
+    Map<String, String> notificationAuth = new HashMap<>();
+    notificationAuth.put("Authorization", "Bearer <notification_token>");
+
     // Initialize the RevAiJobOptions object and assign
     RevAiJobOptions revAiJobOptions = new RevAiJobOptions();
+    // Authorization header is optional; use it if needed to access the media url
```
kbridbur: duplicate comment

examples/AsyncTranscribeMediaUrl.java
```diff
@@ -24,11 +27,26 @@ public static void main(String[] args) {
...
+    revAiJobOptions.setSourceConfig(mediaUrl, sourceAuth);
     revAiJobOptions.setCustomVocabularies(Arrays.asList(customVocabulary));
     revAiJobOptions.setMetadata("My first submission");
-    revAiJobOptions.setCallbackUrl("https://example.com");
+    // Authorization header is optional; use it if needed to access the callback url
```
kbridbur: duplicate comment

examples/CustomVocabularies.java
```diff
@@ -19,18 +20,24 @@ public static void main(String[] args) {
...
-    customVocabularySubmission.setCallbackUrl("https://example.com");
+    customVocabularySubmission.setNotificationConfig(notificationUrl, notificationAuth);
     customVocabularySubmission.setMetadata("My first submission");
 
     CustomVocabularyInformation submittedCustomVocabularyInfo;
 
     try {
       submittedCustomVocabularyInfo =
-          customVocabulariesClient.submitCustomVocabularies(customVocabularySubmission);
+              customVocabulariesClient.submitCustomVocabularies(customVocabularySubmission);
```
kbridbur: Is this too far indented? Lots of stuff in this file appears to have been indented father than it was

src/main/java/ai/rev/speechtotext/models/CustomerUrlData.java
```diff
@@ -0,0 +1,42 @@
+package ai.rev.speechtotext.models;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Map;
+
+public class CustomerUrlData {
```
kbridbur, line 7: Whole class is indented too far. Seems like standard in this repo is 2 spaces indents

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -141,11 +187,40 @@ public String getCallbackUrl() {
...
+    this.notificationConfig = new CustomerUrlData(callbackUrl, authHeaders);
+  }
+
+  /**
+   * Optional property to specify the callback url that Rev AI will POST to when job processing is complete
+   *
+   * @param callbackUrl The url to POST to when job processing is complete.
+   */
+  public void setNotificationConfig(String callbackUrl) {
+    this.notificationConfig = new CustomerUrlData(callbackUrl, null);
```
kbridbur: I think standard in java is to call the overloaded method with some parameters null. This way there's one "source of truth" method and only one place needs to be changed if updates happen.

In this case I don't think it REALLY matters since its just a setter so I'll leave it up to you

kbridbur (commented):
Lots of formatting comments but implementation looks good

README.md
```diff
@@ -11,15 +11,15 @@ See the [API docs](https://docs.rev.ai) for more information about the API.
...
+        <groupId>ai.rev</groupId>
+        <artifactId>revai-java-sdk</artifactId>
+        <version>2.1.0</version>
       </dependency>
 
 ## Build and install locally from source
 Once you've cloned the repo you can use Maven to build it locally and install it in your local Maven .m2 repository.
 
-    mvn install -Dmaven.test.skip=true
+    mvn install -DskipTests=true
```
amikofalvy, line 22: Did the previous instruction not work for you?

examples/AsyncTranscribeMediaUrl.java
aaron-wilson-rev: ~~this is different, I thought it would be good to clarify both the source and callback auth headers are independently optional~~ nvm I see what you mean

src/main/java/ai/rev/speechtotext/models/CustomerUrlData.java
aaron-wilson-rev: oh yeah.... I think that's very non-standard for java but i'll keep it consistent

README.md
aaron-wilson-rev: Yeah I've never seen it with that option before. I get `Unknown lifecycle phase ".test.skip=true"` and I'm on a fresh install of maven. Not sure where it came from though, maybe it works with some other version?

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
aaron-wilson-rev: I can't find a great source on that, but I think it's less elegant to expect the user to have to pass in nulls even if it is a bit harder to support. Most things I'm seeing for optional params in java suggest the overloaded method approach. And since we've only got one optional param something like a builder pattern is overkill imo.

examples/CustomVocabularies.java
aaron-wilson-rev: switched the continuation indent back from 8 to 4

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
kbridbur: I think you misunderstood my comment. I wasn't saying the user should be the one passing nulls. I was saying this method specifically should call setNotificationConfig(callbackUrl, null) instead of duplicating the logic

But as I said before its extremely nitpicky

kbridbur (approved):

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
aaron-wilson-rev: Oh I see what you mean, I completely missed that. Totally agree.

src/main/java/ai/rev/speechtotext/ApiClient.java
```diff
@@ -194,6 +195,7 @@ public String getTranscriptText(String id) throws IOException {
    * @see <a
    *     href="https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob">https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob</a>
    */
+  @Deprecated
```
amikofalvy, line 198: Do we have to deprecate this method? It is a convenience method for submitting a job using the url.

aaron-wilson-rev: We wouldn't have to I guess, but it sets the media url property which is already deprecated. The url only method is still supported. If they want to add both a url and options, I think pointing them to just add the source config to the options makes more sense.

amikofalvy: I was thinking that we can keep the method and change the internals to use source_config instead of media_url property.

This would mean that customers will automatically start using source config. They can use the options method in case they want to provide the more advanced auth header.

amikofalvy: I do see your point that it might be better to deprecate this method. It makes the methods simpler

src/main/java/ai/rev/speechtotext/CustomVocabulariesClient.java
```diff
@@ -42,7 +42,7 @@ public CustomVocabulariesClient(String accessToken) {
    * submission and its progress.
    *
    * @param submission An object that contains the custom vocabularies as well as optional metadata
-   *     and callback url
+   *     and notification config
```
amikofalvy: I think we can remove the enumeration of the internals on the comment for this parameter entirely. No need to list the specifics of this CustomVocabularySubmission object and change it on every addition.

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -15,12 +17,22 @@ public class RevAiJobOptions {
 
   /** The media url where the file can be downloaded. */
   @SerializedName("media_url")
+  @Deprecated
```
amikofalvy, line 22: Is there notation for showing what customers should be using instead?

amikofalvy: If there is please point customers to the new parameter

src/test/java/ai/rev/speechtotext/integration/SubmitJobTest.java
```diff
@@ -122,14 +122,14 @@ public void SubmitJobUrl_UrlAndOptionsAreSpecified_ReturnsRevAiJobInProgress()
...
-    RevAiJob revAiJob = apiClient.submitJobUrl(MEDIA_URL, revAiJobOptions);
+    RevAiJob revAiJob = apiClient.submitJobUrl(revAiJobOptions);
 
     assertRevAiJob(revAiJob);
   }
 
   @Test
   public void SubmitJobUrl_OnlyUrlIsSpecified_ReturnsRevAiJobInProgress() throws IOException {
-    RevAiJob revAiJob = apiClient.submitJobUrl(MEDIA_URL, null);
+    RevAiJob revAiJob = apiClient.submitJobUrl(SOURCE_URL);
```
amikofalvy, line 156: As long as we support the old parameters, even if they are deprecated, I think we should add new tests instead of modifying the existing tests.

Alternatively, you can leave as you have here and add one more tests to check that the old parameter still works

amikofalvy (approved):

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
aaron-wilson-rev: Yeah looks like we should be using "deprecated" in the javadoc alongside this, so I added that as well

aaron-wilson-rev:
Updated more tests for asr and cv to test both the old and new params

PR #43, opened 2022-04-20
AISDK-191: Add topic extraction client

kbridbur:
After doing some research it seems like the http library we use "Retrofit2" does not support generics. The stated reason is that they don't want to handle the complexity of runtime type determination but this doesn't really apply to why they won't handle our use case (we only want generic interfaces, not generic methods). Regardless its not supported. But it seems like the overhead of having an interface per client is pretty minimal so I went ahead with that strategy as opposed to finding a new http package

TODO:
- [x] Tests
- [x] Docs
- [x] Bump major version
- [x] Add an example


src/main/java/ai/rev/exceptions/AuthorizationException.java
```diff
@@ -1,4 +1,4 @@
-package ai.rev.speechtotext.exceptions;
+package ai.rev.exceptions;
```
kbridbur, line 1: I moved a lot of exceptions and helpers that were under speechtotext up a level. This could technically break some users imports so this is going to be a major version bump for the SDK

dmtrrk: would it be reasonable to create a sort of tech debt list for the v2 of the SDK for future, but keep the current version backward compatible?

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,217 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The ApiClient object provides methods to send and retrieve information from all the Rev AI API
```
dmtrrk: is this a correct description of the class? the same is for ctor description

src/main/java/ai/rev/exceptions/AuthorizationException.java
kbridbur: I just bumped the version to v2. I think its somewhat reasonable to do this. It would be significant tech debt to leave it as is

kbridbur: This way when we all any other new apis we only need to do minor version bumps and we don't accrue additional tech debt

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+   * @throws IOException If the response has a status code > 399.
+   * @see TopicExtractionResult
+   */
+  public TopicExtractionResult getResultObject(String id) throws IOException {
+    Map<String, Object> options = new HashMap<>();
+    return apiInterface.getResultObject(id, options).execute().body();
+  }
+
+  /**
+   * The method sends a POST request to the /jobs endpoint, starts an asynchronous job to transcribe
```
Chanc3: The java doc here needs to be updated to this specific job type

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+   * the media file located at the url provided and returns a {@link TopicExtractionJob} object.
+   *
+   * @param text A direct download link to the media.
```
Chanc3: description of the param should be updated

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+      throw new IllegalArgumentException("Job ID must be provided");
+    }
+    apiInterface.deleteJob(id).execute();
+  }
+
+  /**
+   * The method sends a GET request to the /jobs/{id}/transcript endpoint and returns a {@link
+   * TopicExtractionResult} object.
+   *
+   * @param id The ID of the job to return a transcript for.
```
Chanc3: Description needs updating

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+    Map<String, Object> options = new HashMap<>();
+    options.put("threshold", threshold);
+    return apiInterface.getResultObject(id, options).execute().body();
+  }
+
+  /**
+   * The method sends a GET request to the /jobs/{id}/transcript endpoint and returns a {@link
+   * TopicExtractionResult} object.
+   *
+   * @param id The ID of the job to return a transcript for.
```
Chanc3: Description needs updating

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+   * @see TopicExtractionResult
+   */
+  public TopicExtractionResult getResultObject(String id, Double threshold) throws IOException {
+    Map<String, Object> options = new HashMap<>();
+    options.put("threshold", threshold);
+    return apiInterface.getResultObject(id, options).execute().body();
+  }
+
+  /**
+   * The method sends a GET request to the /jobs/{id}/transcript endpoint and returns a {@link
```
Chanc3: Description needs updating

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+    options.setText(text);
+    return apiInterface.submitJob(options).execute().body();
+  }
+
+  /**
+   * An overload of {@link TopicExtractionClient#submitJobText(String, TopicExtractionJobOptions)} without the additional
+   * topic extraction options.
+   *
+   * @param text A direct download link to the media.
```
Chanc3: Description needs updating

src/main/java/ai/rev/topicextraction/TopicExtractionClient.java
```diff
@@ -0,0 +1,250 @@
...
+   * @throws IllegalArgumentException if the text is null.
+   * @see TopicExtractionJob
+   * @see TopicExtractionClient#submitJobText(String, TopicExtractionJobOptions)
+   */
+  public TopicExtractionJob submitJobText(String text) throws IOException {
+    return submitJobText(text, null);
+  }
+  
+  /**
+   * The method sends a POST request to the /jobs endpoint, starts an asynchronous job to transcribe
```
Chanc3: Description needs updating

Chanc3 (commented):
No major requests. Just need to update the java doc descriptions.

examples/TopicExtraction.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    String jobId = submittedJob.getJobId();
+    System.out.println("Job Id: " + jobId);
+    System.out.println("Job Status: " + submittedJob.getJobStatus());
+    System.out.println("Created On: " + submittedJob.getCreatedOn());
+
+    // Waits 5 seconds between each status check to see if job is complete
+    boolean isProcessingComplete = false;
+    while (!isProcessingComplete) {
```
beaudrychase, line 46: Should we add a note that this flow isn't recommended for production?

beaudrychase (approved):
Looks good to me. I had one small comment

examples/TopicExtraction.java
kbridbur: I personally think its fine. The purpose of the example is just for a user to easily be able to see the results of our api

PR #46, opened 2022-05-10
AISDK-190: Update sdks to support notification_config and source_config

aaron-wilson-rev:
Updated topic and sentiment now that they've been added. Also added a couple missing deprecated tags

### DIFF
--- a/src/main/java/ai/rev/sentimentanalysis/models/SentimentAnalysisJobOptions.java
+++ b/src/main/java/ai/rev/sentimentanalysis/models/SentimentAnalysisJobOptions.java
@@ -2,7 +2,6 @@ package ai.rev.sentimentanalysis.models;
 
 import ai.rev.speechtotext.models.CustomerUrlData;
 import ai.rev.speechtotext.models.asynchronous.RevAiTranscript;
-
 import com.google.gson.annotations.SerializedName;
 
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


src/main/java/ai/rev/sentimentanalysis/models/SentimentAnalysisJobOptions.java
```diff
@@ -1,9 +1,12 @@
 package ai.rev.sentimentanalysis.models;
 
+import ai.rev.speechtotext.models.CustomerUrlData;
 import ai.rev.speechtotext.models.asynchronous.RevAiTranscript;
```
dmtrrk: I think we don't need this empty line here.
@kbridbur correct me if I'm wrong, it looks like we have such blocks with all imports ordered inside:
- external libs
- java libs

dmtrrk (approved):

src/main/java/ai/rev/sentimentanalysis/models/SentimentAnalysisJobOptions.java
aaron-wilson-rev: I agree, fixed

dmtrrk: please check other files as well

PR #50, opened 2022-05-23
AISDK-194: Language id should be supported in java sdk

seanlam8:


examples/LanguageIdMediaUrl.java
```diff
@@ -0,0 +1,95 @@
...
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        // Get the language id result
+        LanguageIdResult languageIdResult;
+
+        try {
+            languageIdResult = languageIdClient.getResultObject(jobId);
```
kbridbur: We should print the result is some way to show the user what it looks like

examples/LanguageIdMediaUrl.java
```diff
@@ -0,0 +1,95 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
```
kbridbur: Do you think its worth having some example of how they would take a languageIdResult and use it to submit a transcription job with that language? or is that too simple

src/main/java/ai/rev/languageid/LanguageIdClient.java
```diff
@@ -0,0 +1,313 @@
...
+    public void closeConnection() {
+        client.dispatcher().executorService().shutdown();
+        client.connectionPool().evictAll();
+    }
+
+    /**
+     * This method sends a GET request to the /jobs endpoint and returns a list of {@link LanguageIdJob}
+     * objects.
+     *
+     * @param limit         The maximum number of jobs to return. The default is 100, max is 1000.
```
kbridbur: This is indented to match the other param here but you don't do this elsewhere, I'm not sure which is right but we should be consistent

src/main/java/ai/rev/languageid/LanguageIdClient.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * The method sends a POST request to the /jobs endpoint, starts a language id job for the
+     * provided media url and returns a {@link LanguageIdJob} object.
+     *
+     * @param mediaUrl A direct download link to the media.
+     * @param options  The language id options associated with this job.
+     * @return LanguageIdJob A representation of the language id job.
+     * @throws IOException              If the response has a status code > 399.
```
kbridbur: same thing here

src/main/java/ai/rev/languageid/LanguageIdClient.java
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    private LanguageIdJob submitMultipartRequest(
+            InputStream inputStream, String fileName, LanguageIdJobOptions options) throws IOException {
+        RequestBody fileRequest = FileStreamRequestBody.create(inputStream, MediaType.parse("audio/*"));
+        MultipartBody.Part filePart = MultipartBody.Part.createFormData("media", fileName, fileRequest);
+        return apiInterface.submitJobLocalFile(filePart, options).execute().body();
+    }
+
+
```
kbridbur, line 278: extra newlines

src/main/java/ai/rev/languageid/models/LanguageConfidence.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Represents a predicted language and its corresponding confidences score.
+ */
+public class LanguageConfidence {
+
+    /**
+     * Language code of predicted language.
+     */
+    @SerializedName("language")
+    private String language;
```
kbridbur: getters and setters needed for both properties

src/main/java/ai/rev/languageid/models/LanguageIdJobOptions.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Specifies the source media download url.
+     *
+     * @param sourceMediaUrl The direct download url to the file.
+     */
+
+    public void setSourceConfig(String sourceMediaUrl) {
+        this.sourceConfig = new CustomerUrlData(sourceMediaUrl, null);
```
kbridbur:
```suggestion
        setSourceConfig(sourceMediaUrl, null);
```
Technically this should be done this way so theres a single source of truth method.

src/main/java/ai/rev/languageid/models/LanguageIdJobStatus.java
```diff
@@ -0,0 +1,46 @@
+package ai.rev.languageid.models;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * Specifies constants that define Rev AI job statuses.
```
kbridbur:
```suggestion
 * Specifies constants that define Rev AI language id job statuses.
```

src/main/java/ai/rev/languageid/models/LanguageIdResult.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * A Topic Extraction Result object provides all the information associated with the result of a job.
+ */
+public class LanguageIdResult {
+
+    /**
+     * Language code of the top predicted language.
+     */
+    @SerializedName("top_language")
+    private String topLanguage;
```
kbridbur: needs getter methods

kbridbur (changes requested):
Lots of minor comments, general implementation looks good but we need to fully implement the model classes and be consistent with xml docs spacing

examples/LanguageIdMediaUrl.java
seanlam8: hmmmmmm I think it's pretty simple but I guess it doesn't mean we can't add it.

src/main/java/ai/rev/languageid/models/LanguageIdResult.java
```diff
@@ -0,0 +1,60 @@
+package ai.rev.languageid.models;
+
+import java.util.List;
+
+import ai.rev.topicextraction.models.Topic;
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * A Topic Extraction Result object provides all the information associated with the result of a job.
```
aaron-wilson-rev:
```suggestion
 * A Language Id Result object provides all the information associated with the result of a job.
```

kbridbur (approved):
All the files need to be converted to 2 space indent. They're all 4 right now which is inconsistent with the rest of the SDK. :shipit: after that

PR #53, opened 2022-06-02
AISDK-221 Fix HIPAA Error Message on Incorrect Usage

aaron-wilson-rev:


src/main/java/ai/rev/exceptions/RevAiApiException.java
```diff
@@ -16,36 +16,6 @@ public RevAiApiException(String message, JSONObject errorResponse, int responseC
...
-    if (errorResponse.has("allowed_values")) {
-      stringBuilder.append(System.lineSeparator());
-      stringBuilder.append("Allowed Values: " + errorResponse.get("allowed_values"));
-    }
-    if (errorResponse.has("current_value")) {
-      stringBuilder.append(System.lineSeparator());
-      stringBuilder.append("Current Value: " + errorResponse.get("current_value"));
-    }
-    return stringBuilder.toString();
+    return String.format("Response code: %s, Error: %s, Api response: %s", responseCode, message, errorResponse);
```
aaron-wilson-rev, line 19: We were not logging any exception information before because the response field was "error" and therefore did not get appended to the string.

 I think it makes more sense to log the full response instead of trying to curate the exact message and risk losing information -- as a dev I'd rather see all the info available.

src/main/java/ai/rev/helpers/ErrorInterceptor.java
```diff
@@ -31,6 +32,8 @@ public Response intercept(Interceptor.Chain chain) throws IOException {
           throw new AuthorizationException(errorResponse);
         case 400:
           throw new InvalidParameterException(errorResponse);
+        case 403:
```
amikofalvy, line 35: Maybe we should adjust the default case so we can capture any future error codes we might want to adopt without having to update the SDKs?

aaron-wilson-rev: Changing the string format should make it so the error code and response object show by default. see this comment https://github.com/revdotcom/revai-java-sdk/pull/53/files#r888153910

src/main/java/ai/rev/exceptions/ForbiddenRequestException.java
```diff
@@ -0,0 +1,14 @@
+package ai.rev.exceptions;
+
+import org.json.JSONObject;
+
+/**
+ * The ForbiddenRequestException happens when the request contains parameters that do not pass validation
+ * or the request is not allowed for the user
+ */
+public class ForbiddenRequestException extends RevAiApiException {
```
dmtrrk, line 9: Do we need to add this exception to a Change Log? or adding some example to [Doc](https://github.com/revdotcom/revai-java-sdk/blob/develop/README.md)

src/main/java/ai/rev/helpers/ErrorInterceptor.java
```diff
@@ -31,6 +32,8 @@ public Response intercept(Interceptor.Chain chain) throws IOException {
...
+          throw new ForbiddenRequestException(errorResponse);
```
dmtrrk, line 36: is that possible to create a test for this exception?

eugenep-rev (approved):

src/main/java/ai/rev/exceptions/RevAiApiException.java
dmtrrk: not sure why do we need to have a response code field and API response that contains the response code?
How about:
```
<exception>: <message>\n
<api-response>
```

example:
```
ai.rev.exceptions.ResourceNotFoundException: Resource Not Found Exception
{"extensions":{},"type":"https://www.rev.ai/api/v1/errors/job-not-found","title":"could not find job","status":404}

```

src/main/java/ai/rev/exceptions/ForbiddenRequestException.java
aaron-wilson-rev: Yeah we should add it to the changelog when we run the release. I don't think we need a doc example for this though, it's not like we need to show users how to get this error and it's pretty self-explanatory.

src/main/java/ai/rev/helpers/ErrorInterceptor.java
aaron-wilson-rev: added it to the the test suite

src/main/java/ai/rev/exceptions/RevAiApiException.java
aaron-wilson-rev: The api response code isn't in the error response we're writing out here (it's actually just the response body) so we need to pass the response code separately. IMO we shouldn't do all this extra error handling and formatting at all, but that should be another ticket.

aaron-wilson-rev:
created https://bugs.rev.com/browse/AISDK-223 to cover future testing

src/main/java/ai/rev/exceptions/RevAiApiException.java
dmtrrk: above is the example of the real response I received using your package

aaron-wilson-rev: Not sure how you're getting that. I see
`Exception in thread "main" java.lang.RuntimeException: Failed to submit url [https://www.rev.aidasf/FTC_Sample_1.mp3] Response code: 403, Error: Forbidden Request Exception, Api response: {"error":"HIPAA user is not allowed to use MediaUrl"}
	at ai.rev.AsyncTranscribeMediaUrl.main(AsyncTranscribeMediaUrl.java:65)
`

aaron-wilson-rev: It looks like the status is included with the response body from the API for some errors but not others. I can see it for some other errors but not for the hipaa/media_url validation failure. So I guess it is better to keep the duplicate for now.

PR #54, opened 2022-06-08
AISDK-226 Update Node and Java release version

aaron-wilson-rev:
Changed the version from 2.4.0 I'd set on the feature branch back to 2.3.1 since this is a bug patch release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <!-- groupId, artifactId, these namespaces should not be changed -->
     <groupId>ai.rev</groupId>
     <artifactId>revai-java-sdk</artifactId>
-    <version>2.3.0</version>
+    <version>2.3.1</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
     <!-- groupId, artifactId, these namespaces should not be changed -->
     <groupId>ai.rev</groupId>
     <artifactId>revai-java-sdk</artifactId>
-    <version>2.4.0</version>
+    <version>2.3.1</version>
```
k-weng, line 8: Why did it downgrade?

aaron-wilson-rev: https://github.com/revdotcom/revai-java-sdk/pull/54#issue-1264973238 I changed it to 2.4 in my last pr but it should just be a patch release instead. We never released 2.4.0

k-weng (approved):

PR #55, opened 2022-06-13
AISDK-202: Add Speaker Names option

beaudrychase:


src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -125,6 +125,13 @@ public class RevAiJobOptions {
   @SerializedName("segments_to_transcribe")
   private List<SegmentToTranscribe> segmentsToTranscribe;
 
+  /**
+   * Optional and only available with transcriber "human".
```
amikofalvy, line 129: Can you defer this explanation to the documentation site? I don't want to replicate code that we will need to update if we decide to add this to Ai transcription as well.

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -514,4 +521,23 @@ public List<SegmentToTranscribe> getSegmentsToTranscribe() {
...
+   *
+   * @return List of speaker names
+   */
+  public List<String> getSpeakerNames() {
+    return speakerNames;
+  }
+
+  /**
+   * Specifies the list of speaker names in an audio file
+   * This property is optional but can only be used with "human" transcriber.
```
amikofalvy, line 536: ditto as previous comment. I think we should keep optional in the description, but defer the explanation of human usage only to the API docs site. I don't feel too strongly on this point, so please push back if you think otherwise.

amikofalvy (approved):

amikofalvy (changes requested):
Requesting changes for unit tests.

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
beaudrychase: I guess I didn't consider this because I was just following the established pattern. Are you suggesting to delete the line saying that it's only for 'human' entirely and adding a link to the option in our documentation. I'm hesitant to add links because of the potential for broken links.

amikofalvy: Yeah, up to you. I was just giving another opinion to consider for our docs.
I was trying to optimize for least number of SDK updates to release/ or change how a feature works

beaudrychase: I guess now I'm wondering if there should be anything at all. Should we just delete the 'only for human' lines entirely and not replace them with anything? We don't mention similar restrictions on foreign languages or hipaa

amikofalvy: You could continue with what you have and we can discuss it async at our next office hours

beaudrychase: I'll bring it up at OH

README.md
```diff
@@ -88,6 +88,11 @@ segment.setStartTimestamp(2.0);
 segment.setEndTimestamp(100.5);
 options.setSegmentsToTranscribe(List.of(segment));
 
+// optional speaker names 
+SpeakerName speaker = new SpeakerName();
+speaker.SetDisplayName('Steve');
```
kbridbur:
```suggestion
speaker.setDisplayName('Steve');
```

src/test/java/ai/rev/speechtotext/unit/RevAiJobTest.java
```diff
@@ -171,6 +173,42 @@ public void SubmitJobUrl_UrlAndOptionsAreSpecified_ReturnsARevAiJob() throws IOE
     assertRevAiJob(revAiJob, mockInProgressJob);
   }
 
+  @Test
+  public void SubmitJobUrl_UrlAndHumanTcOptionsSpecified_ReturnsARevAiJob() throws IOException {
```
kbridbur: I'm not sure this needs a separate unit test. The SDK has no knowledge or validation for human vs non human options. We can just keep it to one test where we set all options

beaudrychase: that's a good point, I'll update the tests

kbridbur (changes requested):
SpeakerInfo object needs to be added to the transcript object for this info to actually be used

README.md
```diff
@@ -88,6 +88,11 @@ segment.setStartTimestamp(2.0);
 segment.setEndTimestamp(100.5);
 options.setSegmentsToTranscribe(List.of(segment));
 
+// optional speaker names 
+SpeakerInfo speaker = new SpeakerInfo();
+speaker.setDisplayName('Steve');
```
kbridbur: Middle and last name too please

src/main/java/ai/rev/speechtotext/models/asynchronous/RevAiJobOptions.java
```diff
@@ -514,4 +521,23 @@ public List<SegmentToTranscribe> getSegmentsToTranscribe() {
   public void setSegmentsToTranscribe(List<SegmentToTranscribe> segmentsToTranscribe) {
     this.segmentsToTranscribe = segmentsToTranscribe;
   }
+
+   /**
+   * Returns the list of speaker names
+   *
+   * @return List of speaker names
+   */
+  public List<SpeakerInfo> getSpeakerNames() {
```
kbridbur:
```suggestion
  public List<SpeakerName> getSpeakerNames() {
```

src/main/java/ai/rev/speechtotext/models/asynchronous/SpeakerInfo.java
```diff
@@ -0,0 +1,51 @@
+package ai.rev.speechtotext.models.asynchronous;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * A Segment object represents the information related to the speaker in a transcript
```
kbridbur:
```suggestion
 * A SpeakerInfo object represents the information related to the speaker in a transcript
```

kbridbur (approved):
Seems like you confused the two objects in the request. Other than that looks good. Please verify everything works before merging

PR #56, opened 2022-07-24
AISDK-229: Add skip_postprocessing option

kbridbur:


src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientOptionalParametersTest.java
```diff
@@ -48,17 +48,17 @@ public static String[] vocabIdAndNull() {
...
   @DataPoints("deleteAfterSecondsAndNull")
   public static Integer[] deleteAfterSecondsAndNull() {
-    return new Integer[] {0, 1, null};
+    return new Integer[] {1, null};
   }
 
   @DataPoints("booleanValuesAndNull")
   public static Boolean[] booleanValuesAndNull() {
-    return new Boolean[] {true, false, null};
+    return new Boolean[] {false, null};
```
beaudrychase, line 56: I'm a little confused by this change. Why isn't true a value here

beaudrychase (approved):
I had one question but it looks good to me

src/test/java/ai/rev/speechtotext/unit/RevAiStreamingClientOptionalParametersTest.java
kbridbur: This test had too many possibilities (testing machine was running out of memory) so i got rid of unnecessary options to reduce it. the purpose of the test is to make sure null values don't appear in the query string, so testing anything other than 1 null and 1 nonnull is unneeded

beaudrychase: makes sense

PR #62, opened 2023-12-26
REVAI-3854: Usage examples

kirillatrev:
REVAI-3854: Usage examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,7 @@ try {
 }
 RevAiJob revAiJob = apiClient.submitJobLocalFile(fileInputStream, String fileName, RevAiJobOptions options);
 ```
-
 You can request transcript summary.
-
 ```
 String urlLinkToFile = "https://www.rev.ai/FTC_Sample_1.mp3";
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -69,6 +69,33 @@ try {
...
+You can request transcript summary.
+
+```
+String urlLinkToFile = "https://www.rev.ai/FTC_Sample_1.mp3";
+
+RevAiJobOptions revAiJobOptions = new RevAiJobOptions();
+revAiJobOptions.setSourceConfig(urlLinkToFile, null);
+revAiJobOptions.setLanguage("en");
+revAiJobOptions.setSummarizationOptions(new SummarizationOptions().setModel(NlpModel.STANDARD));
+
```
dmtrrk: redundant
upd. the line 81 is the empty line. it does not look like we need it

examples/AsyncSummarizeMediaUrl.java
```diff
@@ -0,0 +1,113 @@
...
+      throw new RuntimeException("Failed to submit url [" + mediaUrl + "] " + e.getMessage());
+    }
+    String jobId = submittedJob.getJobId();
+    System.out.println("Job Id: " + jobId);
+    System.out.println("Job Status: " + submittedJob.getJobStatus());
+    System.out.println("Created On: " + submittedJob.getCreatedOn());
+
+    // Waits 5 seconds between each status check to see if job is complete
+    boolean isJobComplete = false;
+    while (!isJobComplete) {
```
dmtrrk, line 48: once this is a summarization example, not sure we need this block as it makes the example more complicated.
can we do a summarization polling only?

kirillatrev: @alexsku are you okay with that?
You were requesting polling those jobs in order in python sdk...

dmtrrk (approved):
there is a question regarding polling, but it looks like this approach is ok for SDKs in other languages

PR #63, opened 2024-01-03
REVAI-3919:Split NlpModel to have separate enums for Trans and Sum

kirillatrev:
Split NlpModel to have separate enums for Translation and Summarization

https://revinc.atlassian.net/browse/REVAI-3919

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <!-- groupId, artifactId, these namespaces should not be changed -->
     <groupId>ai.rev</groupId>
     <artifactId>revai-java-sdk</artifactId>
-    <version>2.4.0</version>
+    <version>2.4.1</version>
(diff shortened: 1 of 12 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
     <!-- groupId, artifactId, these namespaces should not be changed -->
     <groupId>ai.rev</groupId>
     <artifactId>revai-java-sdk</artifactId>
-    <version>2.4.0</version>
+    <version>2.4.1</version>
```
dmtrrk, line 8: I think [this](https://semver.org/#summary) is relevant to the change

not sure whether it was discussed.
as a fact, this is a breaking change so it should be `3.0.0`.
but technically we have not announced/documented this feature yet.
how about increasing a minor version?

alexsku (approved):
:shipit:

